PR #33, opened 2022-03-24
WIP create a new xpub from a token, without the admin key

icellan:


mrz1836:
@icellan whats the status on this PR?

PR #108, opened 2022-07-19
docker-compose missing healtchecks

Klar:
I was unable to start the docker-compose as healthchecks are missing.

Added healthchecks, maybe there are some better ones but they work.


mrz1836 (changes requested):
@Klar  Failed YAML linter:

`Error: r-compose.yml:22:33: [error] too few spaces after comma (commas)`

PR #152, opened 2022-09-22
Return transaction from DB when recording a transaction that is already inserted

icellan:
Bux doesn't care who inserts a transaction and no need to trow an error here.


mrz1836 (approved):
LGTM

PR #155, opened 2022-09-29
WIP console integrated into server

icellan:


galt-tr:
React 18 is incompatible with `@mui/styles`, so either need to downgrade to react 17 (tested that and it works) or remove it from `src/components/json-view`.

PR #305, opened 2023-09-20
feat(BUX-234): delete ProcessMempoolOnConnect from config

vitalibalashka:
Screenshot 2023-09-20 at 08 50 17
ProcessMempoolOnConnect was deleted in bux in https://github.com/BuxOrg/bux/pull/401.
So we need to delete it from bux-server too


mrz1836 (approved):
LGTM

PR #366, opened 2023-12-08
feat(BUX-361): bux-server default config

kuba-4chain:
# Pull Request Checklist

# What has changed

### 2. Default values can be overridden by (in this order):
1. flags
2. ENV variables
3. config file

Overriding happens by specifying config path from the config struct. Let's take this as an example:
```go
// AppConfig is the configuration values and associated env vars
type AppConfig struct {
	Authentication     *AuthenticationConfig `json:"auth" mapstructure:"auth"`
	Beef               *BeefConfig           `json:"beef" mapstructure:"beef"`
	Cachestore         *CachestoreConfig     `json:"cache" mapstructure:"cache"`
	ClusterConfig      *ClusterConfig        `json:"cluster" mapstructure:"cluster"`
	...
}

// AuthenticationConfig is the configuration for Authentication
type AuthenticationConfig struct {
	AdminKey        string `json:"admin_key" mapstructure:"admin_key"`               // key that is used for administrative requests
	RequireSigning  bool   `json:"require_signing" mapstructure:"require_signing"`   // if the signing is required
	Scheme          string `json:"scheme" mapstructure:"scheme"`                     // authentication scheme to use (default is: xpub)
	SigningDisabled bool   `json:"signing_disabled" mapstructure:"signing_disabled"` // NOTE: Only for development (turns off signing)
}

// Authentication config option keys for Viper
const (
	AuthAdminKey           = "auth.admin_key"
	AuthRequireSigningKey  = "auth.require_signing"
	AuthSchemeKey          = "auth.scheme"
	AuthSigningDisabledKey = "auth.signing_disabled"
)
```

The path to Admin Key for Authentication is: auth.admin_key. To change it with flags, specify:
```bash
go run ./cmd/server/main --auth.admin_key key123
```
To change with ENVs, replace every dot "." with underscore "\_" + specify "BUX\_" prefix + all uppercase:
```bash
BUX_AUTH_ADMIN_KEY=key123
```

### 3. Environment files are gone. 
There is only one config file now: **config.json**. If this file exists in the **root folder**, it is loaded by default. If it doesn't exist, default values are used.

To change the config file path, use:
```bash
go run ./cmd/server/main.go -C /my/custom/config.json
```
or:
```bash
BUX_CONFIG_FILE="/my/custom/config.json go run ./cmd/server/main.go
```

To generate (dump) config file, with default values, use:
```bash
go run ./cmd/server/main.go -d
```
or combine with -c:
```bash
go run ./cmd/server/main.go -d -C /dump/here/config.json
```

# Important notes
One thing that cannot be overridden by flags and/or ENVs, because of its complicated structure is `MinercraftCustomAPIs`. It can only be overridden in a config.json file.

# Things to discuss
 - sane default values
 - which values to completely remove

Candidates for removal:
- monitor
- mongo
- graphql


config/defaults.go
```diff
@@ -0,0 +1,308 @@
...
+const (
+	AuthAdminKeyDefault        = "xpub661MyMwAqRbcFaYeQLxmExXvTCjw9jjBRpifkoGggkAitXNNjva4TStLJuYjjEmU4AzXRPGwoECjXo3Rgqg8zQqW6UPVfkKtsrogGBw8xz7"
+	AuthRequireSigningDefault  = false
+	AuthSchemeDefault          = "xpub"
+	AuthSigningDisabledDefault = true
+)
+
+// Beef defaults
+const (
+	UseBeefDefault                  = true
```
wregulski, line 29: I would switch UseBeefDefault to false, because if someone uses Beef he/she needs to set Pulse endpoint also. It means that the default value (true) without changing few options in this section won't work properly.

config/defaults.go
```diff
@@ -0,0 +1,308 @@
...
+)
+
+// Nodes defaults
+const (
+	NodesUseMapiFeeQuotesDefault = true
+	NodesMinercraftAPIDefault    = "mAPI"
+)
+
+// NodesBroadcastClientAPIsDefault is the default template of broadcast-client api tokens
+var NodesBroadcastClientAPIsDefault = []string{"url|token"}
```
wregulski, line 118: I'm not sure (please check) but using "url|token" was enabling go-broadcast client and it was trying to call for query/broadcast using:
`url/tx/xyz Bearer: token`
@pawellewandowski98 do you remember?

config/flags.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+type cliFlags struct {
+	showVersion bool `mapstructure:"version"`
+	showHelp    bool `mapstructure:"help"`
+	dumpConfig  bool `mapstructure:"dump_config"`
+}
+
+func loadFlags() error {
+	cli := cliFlags{}
```
wregulski, line 22: Kuba, it looks that you are right. 
I think it is optimized way to serve flags to viper.

I would suggest one improvement - small function like:
```go
func anyFlags() {
   if os.Args > 1 {
    return true
}

return false
}
```

We can call it in method `loadFlags() ` in line 22.

config/defaults.go
```diff
@@ -0,0 +1,308 @@
...
+
+import (
+	"github.com/spf13/viper"
+)
+
+// DefaultConfigFilePath is in the root folder by default
+const DefaultConfigFilePath = "config.json"
+
+// General defaults
+const (
```
arkadiuszos4chain, line 11: From my point of view all these consts are redundant and add unnecessary boilerplate code

config/load.go
```diff
@@ -11,87 +11,68 @@ import (
 	"github.com/spf13/viper"
 )
 
-// isValidEnvironment will return true if the testEnv is a known valid environment
```
arkadiuszos4chain, line 14: <3

config/flags.go
```diff
@@ -0,0 +1,220 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+type buxFlags struct {
```
arkadiuszos4chain, line 11: Why we need this struct? Couldn't we use `pflag.FlagSet` directly?

config/validate.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if err = a.validateDatastore(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// validateDatastore will check the datastore and validate basic requirements
+func (a *AppConfig) validateDatastore() error {
```
arkadiuszos4chain, line 52: Shouldn't be a part of `Datastrore.Validate()` ?

config/defaults.go
```diff
@@ -0,0 +1,308 @@
...
+	SQLReplicaDefault                   = false
+	SQLSkipInitializeWithVersionDefault = true
+	SQLTimeZoneDefault                  = "UTC"
+	SQLTxTimeoutDefault                 = "10s"
+	SQLUserDefault                      = "postgres"
+)
+
+// SQLite config keys
+const (
+	SQLiteDatabasePathDefault = "./test-json.db"
```
dorzepowski, line 79: maybe we should change that name to `bux.db`?

config/defaults.go
```diff
@@ -0,0 +1,308 @@
...
+	SQLiteSharedDefault       = true
+)
+
+// Graphql defaults
+const (
+	GraphqlEnabledDefault        = true
+	GraphqlPlaygroundPathDefault = "/graphql"
+	GraphqlServerPathDefault     = "/graphql"
```
dorzepowski, line 87: Maybe we could consider removing the possibility of setting those GraphQL server and playground paths, and we could just put them statically in a code where it is used? This would mean that in default configuration we will end up with something like this:
```json
"graphql": {
   "enabled": false
}
```
Which is less eye caching and is easier to remove silently later 🙂

config/defaults.go
dorzepowski: Yes, and then there was an error from go-broadcast client that the URL is malformed. 
IMHO we need to talk with BA what should be the default for Arc API url

config/defaults.go
```diff
@@ -0,0 +1,308 @@
...
+	NotificationsEnabledDefault         = false
+	NotificationsWebhookEndpointDefault = ""
+)
+
+// Paymail defaults
+const (
+	PaymailEnabledDefault                 = true
+	PaymailDefaultFromPaymailDefault      = "from@domain.com"
+	PaymailDefaultNoteDefault             = "bux Address Resolution"
+	PaymailDomainValidationEnabledDefault = false
```
dorzepowski, line 131: I'm not sure, but I feal like we should have this enabled by default.

config/defaults.go
```diff
@@ -0,0 +1,308 @@
...
+	RedisURLDefault                   = "redis://localhost:6379"
+	RedisUseTLSDefault                = false
+)
+
+// TaskManager defaults
+const (
+	TaskManagerEngineDefault    = "taskq"
+	TaskManagerFactoryDefault   = "memory"
+	TaskManagerQueueNameDefault = "development_queue"
+)
```
dorzepowski, line 154: Do we know, if there are other options? Maybe we could at least get rid of the Engine and Queue name choices, and just give the possibility to choose between "memory" and "redis" based task manager.

config/load.go
```diff
@@ -11,87 +11,68 @@ import (
...
-	}
+	setDefaults(configFilePath)
 
-	// Get the working directory
-	var workingDirectory string
-	if len(customWorkingDirectory) > 0 {
-		workingDirectory = customWorkingDirectory
-	} else {
-		workingDirectory = getWorkingDirectory()
+	if err = loadFlags(); err != nil {
```
dorzepowski, line 24: As we're setting defaults on a flags definitions, do we still need this setDefault function?

config/validate_server.go
```diff
@@ -1,22 +1,35 @@
...
 func (s *ServerConfig) Validate() error {
-
 	// Set defaults
 	if s.IdleTimeout.String() == "0s" {
-		s.IdleTimeout = DefaultHTTPRequestIdleTimeout
+		duration, err := time.ParseDuration(ServerIdleTimeoutDefault)
+		if err != nil {
+			return err
+		}
+		s.IdleTimeout = duration
```
dorzepowski, line 18: I don't like this idea here.
This is method is called `Validate`, therefore if the s.IdleTImeout is not expected number (for example 0) then we should just return error, instead of setting the default. The other code is responsible for defaults.
The same applies to lines below

config/validate_server_test.go
```diff
@@ -53,8 +54,13 @@ func TestServerConfig_Validate(t *testing.T) {
...
-		assert.Equal(t, DefaultHTTPRequestWriteTimeout, s.WriteTimeout)
-		assert.Equal(t, DefaultHTTPRequestReadTimeout, s.ReadTimeout)
+		idleTimeout, _ := time.ParseDuration(ServerIdleTimeoutDefault)
+		assert.Equal(t, idleTimeout, s.IdleTimeout)
+
+		writeTimeout, _ := time.ParseDuration(ServerWriteTimeoutDefault)
+		assert.Equal(t, writeTimeout, s.WriteTimeout)
+
+		readTimeout, _ := time.ParseDuration(ServerReadTimeoutDefault)
+		assert.Equal(t, readTimeout, s.ReadTimeout)
```
dorzepowski, line 64: This is showing in a bright light how this weird solution in ServerConfig.Validate is making a code confusing. What we're testing? That when we're setting values and making validation of them, we will end up with different values which are the default values 😵‍💫

dorzepowski (commented):
Overall Great job!

config/defaults.go
kuba-4chain: Okay, for now I removed it completely and left an empty array []

config/load.go
```diff
@@ -11,87 +11,68 @@ import (
...
-	return dir
-}
-
 // Added a mutex lock for a race-condition
 var viperLock sync.Mutex
 
-// Load all environment variables
-func Load(customWorkingDirectory string) (_appConfig *AppConfig, err error) {
+// Load all AppConfig
+func Load(configFilePath string) (appConfig *AppConfig, err error) {
```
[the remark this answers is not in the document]
kuba-4chain: No, we don't need. I though it would be nice to be able to set it in code in this function, but you're right that it's useless. Removed.

README.md
```diff
@@ -38,25 +38,117 @@
...
+      --request_logging                          request logging from api routers (rest and graphql) (default true)
+      --auth.admin_key string                    key that is used for administrative requests (default "xpub661MyMwAqRbcFaYeQLxmExXvTCjw9jjBRpifkoGggkAitXNNjva4TStLJuYjjEmU4AzXRPGwoECjXo3Rgqg8zQqW6UPVfkKtsrogGBw8xz7")
+      --auth.require_signing                     require signing
+      --auth.scheme string                       authentication scheme to use (default "xpub")
+      --auth.signing_disabled                    NOTE: Only for development, turns off signing (default true)
+      --beef.use_beef                            enables BEEF transaction format, requires Pulse settings (default true)
+      --beef.pulse.url string                    pulse url for validating merkle roots (default "http://localhost:8000/api/v1/chain/merkleroot/verify")
+      --beef.pulse.auth_token string             authentication token for pulse (default "asd")
+      --cache.engine string                      cache engine: redis, freecache or empty (default "freecache")
+      --cluster.coordinator string               redis or memory (default "redis")
```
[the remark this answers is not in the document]
kuba-4chain: I would even go further and remove cluster-specific redis configuration and just leave the normal redis config and use it for cluster if enabled.

config/defaults.go
kuba-4chain: Why? :(

These constants are used both in viper and in flags, that's why we need them. Plus it's better to see / use / set them like that, instead of looking into viper configuration and/or flags configuration. But that's just my opinion, WDYT?

config/validate.go
kuba-4chain: Good point, moved

README.md
```diff
@@ -38,25 +38,117 @@
...
+
+So for example, to override authentication key for admin, for which a path is `admin.auth_key`, do:
+```BUX_ADMIN_AUTH_KEY="auth123```
+
+To generate config file with defaults, use the --dump flag, or:
+```bash
+go run ./cmd/server/main.go -d
+```
+
+The default config file path is **project root**, and the default file name is **config.json**. This can be overridden by -C flag.
```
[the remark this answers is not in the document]
kuba-4chain: Sure, done

actions/graphql/routes.go
```diff
@@ -91,8 +91,8 @@ func RegisterRoutes(router *apirouter.Router, appConfig *config.AppConfig, servi
 	// Set the POST routes
 	router.HTTPRouter.POST(serverPath, h)
 
-	// only show in development mode
-	if appConfig.Environment == config.EnvironmentDevelopment {
+	// only show in debug mode
+	if appConfig.Debug {
```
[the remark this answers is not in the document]
kuba-4chain: Right, I missed it, thanks

config.example.json
```diff
@@ -1,46 +1,68 @@
...
-  "disable_itc": true,
-  "environment": "development",
-  "gdpr_compliance": false,
-  "request_logging": true,
-  "use_mapi_fee_quotes": true,
-  "minercraft_api": "mAPI",
-  "authentication": {
-    "admin_key": "xpub661MyMwAqRbcFrBJbKwBGCB7d3fr2SaAuXGM95BA62X41m6eW2ehRQGW4xLi9wkEXUGnQZYxVVj4PxXnyrLk7jdqvBAs1Qq9gf6ykMvjR7J",
+  "auth": {
+    "admin_key": "xpub661MyMwAqRbcFaYeQLxmExXvTCjw9jjBRpifkoGggkAitXNNjva4TStLJuYjjEmU4AzXRPGwoECjXo3Rgqg8zQqW6UPVfkKtsrogGBw8xz7",
```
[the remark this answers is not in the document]
kuba-4chain: No idea 😄  I reverted that change.

config.example.json
```diff
@@ -1,46 +1,68 @@
...
+      "url": "http://localhost:8000/api/v1/chain/merkleroot/verify"
+    },
+    "use_beef": true
+  },
   "cache": {
     "engine": "freecache"
   },
-  "_cluster": {
-    "prefix": "bux_cluster_",
+  "cluster": {
```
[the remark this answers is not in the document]
kuba-4chain: Good point, done

config/defaults.go
```diff
@@ -0,0 +1,308 @@
...
+// Notification defaults
+const (
+	NotificationsEnabledDefault         = false
+	NotificationsWebhookEndpointDefault = ""
+)
+
+// Paymail defaults
+const (
+	PaymailEnabledDefault                 = true
+	PaymailDefaultFromPaymailDefault      = "from@domain.com"
```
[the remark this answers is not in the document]
kuba-4chain: From what I understand, it's useless, but it would have to be removed in Bux frst

config/defaults.go
kuba-4chain: Okay, done. I left just TaskManagerFactory option.

config/validate_server_test.go
kuba-4chain: You're absolutely right. Changed.

config/flags.go
kuba-4chain: No, we can't for some reason.
`type buxFlags pflag.FlagSet` doesn't work. No idea why. 

And we need this custom type to generate methods against it.

config/defaults.go
```diff
@@ -0,0 +1,308 @@
...
+	ImportBlockHeadersDefault = ""
+	RequestLoggingDefault     = true
+)
+
+// Authentication defaults
+const (
+	AuthAdminKeyDefault        = "xpub661MyMwAqRbcFaYeQLxmExXvTCjw9jjBRpifkoGggkAitXNNjva4TStLJuYjjEmU4AzXRPGwoECjXo3Rgqg8zQqW6UPVfkKtsrogGBw8xz7"
+	AuthRequireSigningDefault  = false
+	AuthSchemeDefault          = "xpub"
+	AuthSigningDisabledDefault = true
```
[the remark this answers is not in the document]
kuba-4chain: I have no idea, but they are both used kinda differently in Bux. I would remove it in Bux first. WDYT?

config/defaults.go
```diff
@@ -0,0 +1,308 @@
...
+	ClusterCoordinatorDefault         = "redis"
+	ClusterPrefixDefault              = "bux_cluser_"
+	ClusterRedisURLDefault            = "localhost:6379"
+	ClusterRedisMaxIdleTimeoutDefault = "10s"
+	ClusterRedisUseTLSDefault         = false
+)
+
+// Datastore defaults
+const (
+	DatastoreAutoMigrateDefault = true
```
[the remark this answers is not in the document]
kuba-4chain: Okay, removed.

config/load.go
kuba-4chain: In theory that would work.

But in practice, I think that it's not the flags responsibility to set global defaults, because flags are not directly from viper, we just bind them. Viper defaults should be the source of truth, not flag defaults. We may have default values that we don't want to expose as flags.

Using the same defaults in flags and in Viper settings is just a convention I used that I think is the most elegant of all options.

README.md
```diff
@@ -38,25 +38,60 @@
...
+    ...
+}
+
+type AuthenticationConfig struct {
+	AdminKey string `json:"admin_key" mapstructure:"admin_key"`
+	RequireSigning bool `json:"require_signing" mapstructure:"require_signing"`
+	Scheme string `json:"scheme" mapstructure:"scheme"`
+	SigningDisabled bool `json:"signing_disabled" mapstructure:"signing_disabled"`
+}
+```
```
dorzepowski, line 89: Instead of go code, I would rather point to example config file, and write that instruction based on it.

dorzepowski (changes requested):

config/config.go
```diff
@@ -2,273 +2,328 @@
...
-	EnvironmentDevelopment         = "development"
-	EnvironmentKey                 = "BUX_ENVIRONMENT"
-	EnvironmentPrefix              = "bux"
-	EnvironmentProduction          = "production"
-	EnvironmentStaging             = "staging"
-	EnvironmentTest                = "test"
-	HealthRequestPath              = "health"
-	Version                        = "v0.5.16"
+	ApplicationName         = "BuxServer"
+	CurrentMajorVersion     = "v1"
```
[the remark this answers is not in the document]
dorzepowski: this is minor, but IMHO it would be better to rename it
```suggestion
	ApiVersion     = "v1"
```

config/defaults.go
```diff
@@ -0,0 +1,156 @@
...
+	Nodes:              NodesDefault,
+	Notifications:      NotificationDefault,
+	Paymail:            PaymailDefault,
+	RequestLogging:     true,
+	Server:             ServerDefault,
+	TaskManager:        TaskManagerDefault,
+}
+
+// AuthConfigDefault is the default config for AuthenticationConfig
+var AuthConfigDefault = &AuthenticationConfig{
```
dorzepowski, line 31: Do we need it as a public var? do we use it somewhere else then just for initilialisation of DefaultAppConfig?

This ☝️ question also applies to all the variables below.

config/load_test.go
```diff
@@ -1,39 +1,38 @@
...
-		valid := isValidEnvironment(EnvironmentTest)
-		assert.Equal(t, true, valid)
+	t.Run("custom configFilePath overridden by ENV", func(t *testing.T) {
+		// given
+		anotherPath := "anotherPath.yml"
 
-		valid = isValidEnvironment(EnvironmentDevelopment)
-		assert.Equal(t, true, valid)
+		// when
+		os.Setenv("BUX_"+strings.ToUpper(ConfigFilePathKey), anotherPath)
```
dorzepowski, line 28: It is one of the most important variables. Therefore I would suggest to statically type env name. Like: 
```suggestion
		os.Setenv("BUX_CONFIG_FILE", anotherPath)
```
Why?

Because we must treat it like a part of the contract of the bux server, therefore if we change a name for variable we should also change the test, so everyone could see that we're doing something backward incompatible.

And don't forget to also change it in a `// cleanup` section (line 36)

dorzepowski (changes requested):

config/defaults.go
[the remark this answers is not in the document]
dorzepowski: Can you add a task/subtask about checking/removing this option (and other, which I point in comments)?

config/defaults.go
kuba-4chain: Removed :)

README.md
kuba-4chain: Good point, done

config/defaults.go
kuba-4chain: Right, changed to private.

config/load_test.go
kuba-4chain: Makes sense, I updated this file

dorzepowski (approved):

PR #377, opened 2023-12-19
feat(BUX-421): replace router logger

pawellewandowski98:
# Description
Unify logs in http router

# Pull Request Checklist

### DIFF
--- a/actions/middleware.go
+++ b/actions/middleware.go
@@ -1,13 +1,17 @@
 package actions
 
 import (
+	"context"
 	"net/http"
+	"time"
 
 	"github.com/BuxOrg/bux"
 	"github.com/BuxOrg/bux-server/config"
 	"github.com/BuxOrg/bux-server/dictionary"
+	"github.com/gofrs/uuid"
 	"github.com/julienschmidt/httprouter"
 	apirouter "github.com/mrz1836/go-api-router"
+	"github.com/mrz1836/go-parameters"
 )
 
 // Action is the configuration for the actions and related services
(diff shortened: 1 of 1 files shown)


actions/middleware.go
```diff
@@ -101,3 +101,33 @@ func CheckAuthentication(appConfig *config.AppConfig, bux bux.ClientInterface, r
...
+		start := time.Now()
+
+		// Fire the request
+		h(writer, req, ps)
+
+		// Complete the timer and final log
+		elapsed := time.Since(start)
+
+		if a.AppConfig.RequestLogging {
+			a.Services.Logger.Debug().Msgf("%d | %s | %s | %s | %s ", writer.Status, elapsed, req.RemoteAddr, req.Method, req.URL)
```
dorzepowski, line 130: Additionally we could add those params in contextual fields and maybe some additional marker like Str("logger", "http-request")

dorzepowski (changes requested):

actions/middleware.go
pawellewandowski98: So I should leave it in a message and additionally add to context?

actions/middleware.go
```diff
@@ -101,3 +101,33 @@ func CheckAuthentication(appConfig *config.AppConfig, bux bux.ClientInterface, r
...
+		// Start the log (timer)
+		start := time.Now()
+
+		// Fire the request
+		h(writer, req, ps)
+
+		// Complete the timer and final log
+		elapsed := time.Since(start)
+
+		if a.AppConfig.RequestLogging {
```
arkadiuszos4chain, line 129: Do we need `RequestLogging` flag if we log it as DEBUG?

pawellewandowski98: We can remove it If we want, It was still in config after refactoring so I used it

actions/middleware.go
pawellewandowski98: Done, I added all parameters into logger

actions/middleware.go
arkadiuszos4chain: Leave it as it is

arkadiuszos4chain (approved):

Nazarii-4chain (approved):

dorzepowski (approved):

PR #390, opened 2024-01-03
feat(BUX-420): go version and workflows update and config fix for k8s

kuba-4chain:
# Pull Request Checklist


arkadiuszos4chain (approved):

config/config.go
```diff
@@ -54,7 +54,7 @@ type AppConfig struct {
 	// RequestLogging is flag for enabling logging in go-api-router.
 	RequestLogging bool `json:"request_logging" mapstructure:"request_logging"`
 	// Server is a general configuration for bux-server.
-	Server *ServerConfig `json:"server" mapstructure:"server"`
+	Server *ServerConfig `json:"server_config" mapstructure:"server_config"`
```
dorzepowski, line 57: Why?

config/load.go
```diff
@@ -39,6 +40,15 @@ func Load(logger *zerolog.Logger) (appConfig *AppConfig, err error) {
 		return nil, err
 	}
 
+	if appConfig.Debug {
+		cfg, err := json.MarshalIndent(appConfig, "", "  ")
+		if err != nil {
+			logger.Error().Msg("Unable to decode App Config to json")
+		} else {
+			fmt.Printf("loaded config: %s", cfg)
+		}
```
dorzepowski, line 49: 1. as we have a struct with config, why don't you just use `Printf("loaded config: %v", appConfig)`?
2. Why you're doing fmt.Printf instead of using a logger?

config/validate_server.go
```diff
@@ -20,10 +19,14 @@ func (s *ServerConfig) Validate() error {
 		return errors.New("Write timeout needs to be set for server configuration")
 	}
 
+	if s.Port < 10 || s.Port > 999999 {
```
dorzepowski, line 22: Still out of bounds:
max port number is 65 535

dorzepowski (changes requested):
I can live with that, but I just want to mention that it isn't best practice to mix bumping go version and changes in config in one PR.

config/config.go
kuba-4chain: This is actually a funny case.

Because we changed the config file from json to yaml, k8s somehow interfered with it and put it's own values into server.port, which broke the server and it couldn't start.

config/load.go
kuba-4chain: 1. Because it will print it in one-liner.
2. Because it will print it in one-liner 😄

Doing it this way, we print it as a formatted json which makes it MUCH easier to look at quickly and seeing if everything is okay. This is also shown only when when `debug` flag is selected. That's why I decided to make an exception here and use the standard fmt.Printf function.

config/validate_server.go
kuba-4chain: You're right, changed

dorzepowski (approved):

PR #425, opened 2024-01-26
BUX-487: AdminGetAccessKeys

Nazarii-4chain:
Fix method not allowed
# Pull Request Checklist


actions/admin/routes.go
```diff
@@ -23,7 +23,7 @@ func RegisterRoutes(router *apirouter.Router, appConfig *config.AppConfig, servi
 	// V1 Requests
 	router.HTTPRouter.GET("/"+config.APIVersion+"/admin/stats", action.Request(router, require.Wrap(action.stats)))
 	router.HTTPRouter.GET("/"+config.APIVersion+"/admin/status", action.Request(router, require.Wrap(action.status)))
-	router.HTTPRouter.POST("/"+config.APIVersion+"/admin/access-keys/search", action.Request(router, require.Wrap(action.accessKeysSearch)))
+	router.HTTPRouter.GET("/"+config.APIVersion+"/admin/access-keys/search", action.Request(router, require.Wrap(action.accessKeysSearch)))
```
chris-4chain, line 26: I'm just wondering, whether the previous `POST` was better idea - because,  as I see it, the request accepts json body - and it suits more to POST than to GET method which should be simple request.  

Maybe the problem is on the other side? Just thinking.

Nazarii-4chain: Requests were made by copy-paste method. The REST API rule is that `GET` should be used to get resources and `POST` for creating or submitting data for processing.

chris-4chain: Ok, I won't argue :)

dorzepowski: I must disagree, unfortunately currently there isn't HTTP method for "search", 
and often POST is better, because it accepts body where you can put much longer "queries" then in url.
Although there is no limitation for URL length in specification, I saw that some network components could reject URL that is to long. 
Therefore I'm voting for using POST in each search request and GET if this is get by id (or other single identifier).

dorzepowski (changes requested):

PR #436, opened 2024-01-30
BUX-493: Admin get transactions

Nazarii-4chain:
# Pull Request Checklist
Changed method to GET


chris-4chain (approved):

actions/admin/routes.go
```diff
@@ -34,7 +34,7 @@ func RegisterRoutes(router *apirouter.Router, appConfig *config.AppConfig, servi
 	router.HTTPRouter.POST("/"+config.APIVersion+"/admin/paymails/count", action.Request(router, require.Wrap(action.paymailAddressesCount)))
 	router.HTTPRouter.POST("/"+config.APIVersion+"/admin/paymail/create", action.Request(router, require.Wrap(action.paymailCreateAddress)))
 	router.HTTPRouter.DELETE("/"+config.APIVersion+"/admin/paymail/delete", action.Request(router, require.Wrap(action.paymailDeleteAddress)))
-	router.HTTPRouter.POST("/"+config.APIVersion+"/admin/transactions/search", action.Request(router, require.Wrap(action.transactionsSearch)))
+	router.HTTPRouter.GET("/"+config.APIVersion+"/admin/transactions/search", action.Request(router, require.Wrap(action.transactionsSearch)))
```
dorzepowski, line 37: Unfortunately currently there isn't HTTP method for "search",
and often POST is better, because it accepts body where you can put much longer "queries" then in url.
Although there is no limitation for URL length in specification, I saw that some network components could reject URL that is to long.
Therefore I'm voting for using POST in each search request and GET if this is get by id (or other single identifier). And if client is not working, it should be changed in a client, not here.

dorzepowski (changes requested):

actions/admin/routes.go
Nazarii-4chain: I found couple examples of using GET as a search request. 
Google, Elastic, Adobe uses GET requests for search.

https://softwareengineering.stackexchange.com/questions/233164/how-do-searches-fit-into-a-restful-interface
https://www.google.com/search?q=skeleton&tbas=0&tbm=isch&tbs=isz:l,ic:gray,itp:face,qdr:w,imgo:1
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html
https://developer.adobe.com/commerce/webapi/rest/use-rest/performing-searches/

dorzepowski: 1. First of all this API isn't REST API, because there are those verbs in a paths.
2. elastic is providing both GET and POST, so if GET won't work for you, then you can switch to POST,
3. Google is limiting your query by either not allowing to put more in an input, or cutting the query, or responding with error.
4. Introducing this changes, makes the bux-server incompatible with: js-buxclient, go-buxclient, bux-console. So a lot of additional changes.

Just to wrap up. It's not like I'm against fixing the design of the API. But:
1. It is out of scope for v1
2. We should plan how to do that - "what if someone will prepare a really big query to search transactions and the infrastructure will reject it" - maybe we should support both GET and POST?
3. We should make a complex adjustment if we want the API to be RESTful.

Nazarii-4chain: I see. Will change methods in clients

PR #439, opened 2024-01-31
feat(BUX-368): Arc callbacks

wregulski:
### !It needs the newest version of Bux to work!

# Pull Request Checklist


pawellewandowski98 (approved):

actions/middleware.go
```diff
@@ -102,6 +113,22 @@ func CheckAuthentication(appConfig *config.AppConfig, bux bux.ClientInterface, r
...
 }
 
+func VerifyCallbackToken(appConfig *config.AppConfig, req *http.Request) (*http.Request, dictionary.ErrorMessage) {
+	const BEARER_SCHEMA = "Bearer "
+	authHeader := req.Header.Get("Authorization")
+	if authHeader == "" {
+		return req, dictionary.GetError(dictionary.ErrorAuthenticationCallback, "missing auth header")
+	}
+
+	providedToken := authHeader[len(BEARER_SCHEMA):]
```
chris-4chain, line 123: Potential Index Out of Range Error: If `authHeader` is shorter than "Bearer ", the slicing operation `authHeader[len(BEARER_SCHEMA):]` could cause an index out of range error.

actions/transactions/broadcastCallback.go
```diff
@@ -0,0 +1,42 @@
...
+		apirouter.ReturnResponse(w, req, http.StatusExpectationFailed, err.Error())
+		return
+	}
+
+	defer req.Body.Close()
+
+	err = a.Services.Bux.UpdateTransaction(req.Context(), resp)
+	if err != nil {
+		msg := fmt.Sprintf("failed to update transaction - tx: %v", resp)
+		a.Services.Logger.Err(err).Msg(msg)
```
chris-4chain, line 35: Why don't use `Msgf` functions?

actions/transactions/broadcastCallback.go
```diff
@@ -0,0 +1,42 @@
...
+		apirouter.ReturnResponse(w, req, http.StatusOK, "")
```
chris-4chain, line 36: Are you sure to return `StatusOK` in case of the error?

actions/middleware.go
chris-4chain: Another, maybe reasonable (or not) issue is that: 
- What if somebody put different string at the beginning of the authorization header with the same length as "Bearer"?This code will still accept this header, even if it's another auth-scheme.

actions/transactions/broadcastCallback.go
wregulski: I didn't know about this function

actions/transactions/broadcastCallback.go
wregulski: Yes, we always want to response with StatusOK to callback as Arc doesn't handle responses for callback - it's "one way ticket". 
But for the clarication I can add 500 or something.

actions/transactions/broadcastCallback.go
wregulski: fixed

actions/transactions/broadcastCallback.go
wregulski: fixed

actions/middleware.go
wregulski: fixed

config/defaults.go
```diff
@@ -56,6 +61,18 @@ func getCacheDefaults() *CacheConfig {
 	}
 }
 
+func getCallbackDefaults(logger *zerolog.Logger) *CallbackConfig {
```
dorzepowski, line 64: Maybe I'm wrong, but actually this code will always generate a callback token from default admin xpub. So by default everyone will have the same token. 
IMHO it should be generated for provided admin xpub.
In the other comment I'm pointing better place to calculate it.
(And this way, we can remove the dependency to logger 🙂

config/services.go
```diff
@@ -183,6 +183,8 @@ func (s *AppServices) loadBux(ctx context.Context, appConfig *AppConfig, testMod
 		options = loadBroadcastClientArc(appConfig, options, logger)
 	}
 
+	options = append(options, bux.WithCallback(appConfig.Callback.CallbackHost+BroadcastCallbackRoute, appConfig.Callback.CallbackToken))
```
dorzepowski, line 186: Maybe this would be a good place to check if CallbackToken is not empty. If not, then we can generate it from admin xpub here.

dorzepowski (changes requested):

config/services.go
wregulski: done

config/defaults.go
wregulski: done

Nazarii-4chain (approved):

PR #479, opened 2024-02-27
BUX-623/ Postgresql engine

Nazarii-4chain:
# Pull Request Checklist
Changed postgres to postgresql, mark mongodb as experimental

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -28,11 +28,10 @@ db:
   datastore:
     # enable datastore debug mode
     debug: false
-    # datastore engine - sqlite/postgresql/mysql/mongodb
-    engine: postgres
+    # datastore engine - sqlite/postgresql/mysql/mongodb (experimental)
+    engine: postgresql
     # prefix for all tables
     table_prefix: xapi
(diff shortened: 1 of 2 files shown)


config.example.yaml
```diff
@@ -28,10 +28,11 @@ db:
   datastore:
     # enable datastore debug mode
     debug: false
-    # datastore engine - sqlite/postgres/mysql/mongodb
-    engine: sqlite
+    # datastore engine - sqlite/postgresql/mysql/mongodb
+    engine: postgres
```
wregulski, line 32: I think it's not correct - IMO it should be:
    # datastore engine - sqlite/postgresql/mysql/mongodb (experimental)
    engine: postgresql

wregulski (changes requested):

config.example.yaml
wregulski: We should still have sqlite as default option

Nazarii-4chain: Thanks. Will fix

wregulski (approved):

PR #490, opened 2024-03-04
Bux-398/ get waiting contacts

Nazarii-4chain:
Added GetContacts action, 
removed unused files, 
change to ShouldBindJson method in contact creation, 
Optimized newContact arguments, 
adjusted ContactService interface, 
changed id generation in contact model, 
added tests, 
added getContacts handler
# Pull Request Checklist

### DIFF
--- a/actions/contacts/create.go
+++ b/actions/contacts/create.go
@@ -24,9 +24,9 @@ import (
 // @Security	bux-auth-xpub
 func (a *Action) create(c *gin.Context) {
 
-	var contactStruct engine.Contact
+	var contact engine.Contact
 
-	err := c.ShouldBindJSON(&contactStruct)
+	err := c.ShouldBindJSON(&contact)
 
 	if err != nil {
 		c.JSON(http.StatusUnprocessableEntity, err.Error())
@@ -37,7 +37,7 @@ func (a *Action) create(c *gin.Context) {
 
 	var requestBody CreateContact
 
-	newContact, err := a.Services.SpvWalletEngine.NewContact(c.Request.Context(), contactStruct.FullName, contactStruct.Paymail, pubKey, engine.WithMetadatas(requestBody.Metadata))
+	newContact, err := a.Services.SpvWalletEngine.NewContact(c.Request.Context(), contact.FullName, contact.Paymail, pubKey, engine.WithMetadatas(requestBody.Metadata))
 	if err != nil {
 		c.JSON(http.StatusUnprocessableEntity, err.Error())
 		return
--- a/actions/contacts/create.go
+++ b/actions/contacts/create.go
@@ -1,10 +1,12 @@
 package contacts
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/bitcoin-sv/spv-wallet/engine"
 	"github.com/bitcoin-sv/spv-wallet/mappings"
+	"github.com/bitcoin-sv/spv-wallet/server/auth"
 	"github.com/gin-gonic/gin"
 )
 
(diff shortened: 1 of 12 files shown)


actions/contacts/create.go
```diff
@@ -23,13 +25,19 @@ import (
...
-	fullName := c.GetString("full_name")
-	paymail := c.GetString("paymail")
-	pubKey := c.GetString("pubKey")
+	var contactStruct engine.Contact
+
+	err := c.ShouldBindJSON(&contactStruct)
+
+	if err != nil {
+		fmt.Println("err", err)
+	}
```
chris-4chain, line 34: I think, fmt.Println should be replaced to use a logger or maybe the printing is not necessary here?

actions/contacts/create.go
```diff
@@ -23,13 +25,19 @@ import (
 // @Security	bux-auth-xpub
 func (a *Action) create(c *gin.Context) {
 
-	fullName := c.GetString("full_name")
-	paymail := c.GetString("paymail")
-	pubKey := c.GetString("pubKey")
+	var contactStruct engine.Contact
```
chris-4chain, line 28: Consider another name without "Struct" suffix.

actions/contacts/create.go
Nazarii-4chain: Changed

actions/contacts/create.go
Nazarii-4chain: Changed

engine/model_contact_test.go
```diff
@@ -102,3 +96,62 @@ func Test_getContact(t *testing.T) {
...
+		defer deferMe()
+
+		var metadata *Metadata
+
+		dbConditions := make(map[string]interface{})
+
+		var queryParams *datastore.QueryParams
+
+		(dbConditions)[xPubIDField] = xPubID
+		(dbConditions)[contactStatus] = notConfirmed
```
chris-4chain, line 112: Is it necessary to put the `dbConditions` variable into parenthesis?

Nazarii-4chain: Changed. Thanks

actions/contacts/create.go
chris-4chain: I can't see that you've changed it :sweat_smile:

Nazarii-4chain: Changed in another file LoL

actions/contacts/search.go
```diff
@@ -0,0 +1,64 @@
...
+// @Produce		json
+// @Param		page query int false "page"
+// @Param		page_size query int false "page_size"
+// @Param		order_by_field query string false "order_by_field"
+// @Param		sort_direction query string false "sort_direction"
+// @Param		conditions query string false "conditions"
+// @Success		200
+// @Router		/v1/contacts [get]
+// @Security	x-auth-xpub
+func (a *Action) search(c *gin.Context) {
```
chris-4chain, line 28: 1. Will this `search` support querying for substrings for a autosuggestion feature?

image

2. Does it support fetching all the contacts at once? - maybe if I don't provide page & page_size params?
3. Is the searching by params case sensitive? I'm wondering what should happen when a user have a contact "BoB@b.com" and he puts into paymail field "bob@b.com". I suppose the app should inform a user that it is a trusted contact.
image

arkadiuszos4chain: 1. Good question. In my opinion, we should be consistent with other `/search` endpoints in the system for now.
2. Leave `page` and `page_size` as optional parameters for consistency, but DO NOT set any MAX_PAGE_SIZE on server side for now.
3. We sanitize paymails (at least we should, I'll ensure we do it during adding contact), so we should make sure the paymail condition is in lower case to not rely on database collation.

Nazarii-4chain: > 1. Will this `search` support querying for substrings for a autosuggestion feature?
> 
> image
> 
> 2. Does it support fetching all the contacts at once? - maybe if I don't provide page & page_size params?
> 3. Is the searching by params case sensitive? I'm wondering what should happen when a user have a contact "[BoB@b.com](mailto:BoB@b.com)" and he puts into paymail field "[bob@b.com](mailto:bob@b.com)". I suppose the app should inform a user that it is a trusted contact. Or maybe the "paymail" should be a `special` field and it should always be transformed to lower case.
>    image

1. It should work. It has the same logic under the hood as other `search`s. 
2. Yes. If you don't provide any metadata and params it should fetch all contacts.
3. Will add sanitizing paymails too

engine/model_contact_test.go
chris-4chain: During review I've noticed that you haven'tt changed it in all places where the `(dbConditions)[` occurs. Please adjust.

chris-4chain (approved):

PR #496, opened 2024-03-07
feat(BUX-641) Shared Config

chris-4chain:
It is a feature which exposes a `SharedConfig` with elements from `Config`. 

image

# Pull Request Checklist

### DIFF
--- a/models/shared_config.go
+++ b/models/shared_config.go
@@ -1,0 +1,13 @@
+package models
+
+// SharedConfig with fields which can ba shared across the application components.
+// Please be aware NOT to add ANY SENSITIVE information here.
+type SharedConfig struct {
+	PaymilDomains        []string           `json:"domains"`
+	ExperimentalFeatures ExperimentalConfig `json:"experimental_features"`
+}
+
+// ExperimentalConfig represents a feature flag config.
+type ExperimentalConfig struct {
+	PikeEnabled bool `json:"pike_enabled" mapstructure:"pike_enabled"`
+}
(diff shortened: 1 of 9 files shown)


models/shared_config.go
```diff
@@ -0,0 +1,13 @@
...
+// SharedConfig with fields which can ba shared across the application components.
+// Please be aware NOT to add ANY SENSITIVE information here.
+type SharedConfig struct {
+	PaymilDomains        []string           `json:"paymail_domains"`
+	ExperimentalFeatures ExperimentalConfig `json:"experimental_features"`
+}
+
+// ExperimentalConfig represents a feature flag config.
+type ExperimentalConfig struct {
+	PikeEnabled bool `json:"pike_enabled" mapstructure:"pike_enabled"`
```
dorzepowski, line 12: Maybe it's worth to have this solution more flexible.
I mean, in the current implementation everytime we add an experimental feature we need to update go-client also to use it in the wallet.
If we would expose a simple key=>boolean map on an endpoint, then actually we could just check if given string is there and is true. 

It is only on endpoint. In spv-wallet I would rather keep it somehow more typed so we can easier use it arround the code and easier remove it.

dorzepowski (changes requested):

models/shared_config.go
chris-4chain: I've modified my solution. From now on, the `ExperimentalConfig` struct is defined in `config` package (as it used to be) and the `model.SharedConfig` has a `map[string]bool` for ExperimentalConfig fields. 

Also adjusted `-web-backend` and then tested locally.

dorzepowski (approved):

PR #515, opened 2024-03-25
feat(bux-685): sanitized and parse fields

Nazarii-4chain:
# Pull Request Checklist


engine/model_destinations.go
```diff
@@ -163,11 +163,38 @@ func getDestinationsByXpubID(ctx context.Context, xPubID string, usingMetadata *
...
+			sanitizedValue := utils.SanitizeInput(value.(string))
+			if _, ok := dbConditions[key]; ok {
+				dbConditions[key] = sanitizedValue
+			}
+
+			return nil, fmt.Errorf("column does not exist: %s", key)
+		}
 	}
 	dbConditions[xPubIDField] = xPubID
+	//result[xPubIDField] = xPubID
```
chris-4chain, line 197: General thoughts:
- JSON naming (by struct tags) doesn't equal with GORM naming strategy, so you cannot just parse a model to a json and back to get appropriate keys. Check out the solution I've found:

```golang
"gorm.io/gorm/schema"
// (...)
sch, err := schema.Parse(theModelYouWantToParse{}, &sync.Map{}, schema.NamingStrategy{})
```

More about naming strategy: https://gorm.io/docs/conventions.html

- Values of filters don't have to be sanitized because they are sql-injection safe - the problem is with the keys. 
- In your current implementation, I guess, even though an user doesn't provide any conditions the `dbConditions` are filled with empty strings which may lead to wrong behaviour. 

---
Generally, I'm thinking that maybe we could consider different approach, like that:
Just a golang-like pseudocde:
```golang
type FilterDestinations struct {
   ID         *string
   Type     *string
   DraftID *string
}

func getDestinationsByXpubID(/* other args */, filter *FilterDestinations) {
   conditions:= map[string]interface{}{
      xPubIDField: xPubID
   }
   if filter != nil {
      if filter.ID != nil {
          conditions["id"] = filter.ID
      }
      //apply other conditions - can be shtortened by some util function
   }
}
```

Nazarii-4chain: Thanks. Will have a look on this

PR #519, opened 2024-03-28
feat(bux-685): (Sql Injection) Added DestinationFilter struct, added filter parser

Nazarii-4chain:
# Pull Request Checklist


engine/utils/parseFilters.go
```diff
@@ -0,0 +1,41 @@
+package utils
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+// FilterMapByStructFields returns a new map containing only the fields that exist in the provided struct.
+func FilterMapByStructFields(fieldMap map[string]interface{}, filtersStruct interface{}) (map[string]interface{}, error) {
```
dorzepowski, line 10: I have few concerns about this function:
1. It can't handle date ranges and custom flags like "IncludeDeleted"
2. It has the same issue as with your previous PR, that actually gorm names are different then json names (and actually in the future if we would like to switch to camelcase on json it would make it even more different)
3. there are no tests that would show what is the expected behaviour of this function and if it is really doing what it meant to do.


So from my point of view I would go with something more straight forward like something what @chris-4chain suggested in previous PR. So something like:
```
conditions := destFilters.ToConditions()

// ....

func (d *DestinationFilters) ToConditions() {

if d.LockingScript != nil {
  addToConditions(conditions, "locking_script", d.LockingScript)
}

//...

```

actions/destinations/search.go
```diff
@@ -32,12 +33,21 @@ func (a *Action) search(c *gin.Context) {
 		return
 	}
 
+	var destFilters models.DestinationFilters
```
dorzepowski, line 36: This model should be extracted from request body instead of conditions map.

Also it should be included in Swagger

models/destination_filters.go
```diff
@@ -0,0 +1,22 @@
...
+	// LockingScript is a destination's locking script.
+	LockingScript *string `json:"locking_script,omitempty" example:"76a9147b05764a97f3b4b981471492aa703b188e45979b88ac"`
+	// Address is a destination's address.
+	Address *string `json:"address,omitempty" example:"1CDUf7CKu8ocTTkhcYUbq75t14Ft168K65"`
+	// DraftID is a destination's draft id.
+	DraftID *string `json:"draft_id,omitempty" example:"b356f7fa00cd3f20cce6c21d704cd13e871d28d714a5ebd0532f5a0e0cde63f7"`
+	// IncludeDeleted is a flag which includes deleted destinations.
+	IncludeDeleted *bool `json:"include_deleted,omitempty" example:"true"`
+	// Metadata is a metadata map of outer model.
+	Metadata *map[string]interface{} `json:"metadata,omitempty" swaggertype:"object,string" example:"key:value,key2:value2"`
```
dorzepowski, line 17: I think that it is part of higher struct, the SearchRequestParameters, but we can consider makeing it a part of this one and removing it from enclosing one.

dorzepowski (changes requested):

engine/utils/parseFilters.go
chris-4chain, line 10: It looks like the `FilterMapByStructFields` function is no longer being used after my changes. Could you please remove it?

It would also be great if you could take a quick look for any other functions that might not be needed anymore.

chris-4chain (changes requested):

chris-4chain (approved):

engine/go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/DATA-DOG/go-sqlmock v1.5.2
 	github.com/bitcoin-sv/go-broadcast-client v0.17.2
 	github.com/bitcoin-sv/go-paymail v0.14.0
+	github.com/bitcoin-sv/spv-wallet/models v0.25.0
```
dorzepowski, line 9: Why we have dependency to models from engine? It shouldn't be here, as models is used only at the API layer

engine/go.mod
```diff
@@ -160,3 +161,5 @@ replace github.com/gomodule/redigo => github.com/gomodule/redigo v1.8.9
 // Issue: go.mongodb.org/mongo-driver/x/bsonx: cannot find module providing package go.mongodb.org/mongo-driver/x/bsonx
 // Need to convert bsonx to bsoncore
 replace go.mongodb.org/mongo-driver => go.mongodb.org/mongo-driver v1.11.7
+
+replace github.com/bitcoin-sv/spv-wallet/models => ../models
```
dorzepowski, line 165: IMHO it is not needed

dorzepowski (changes requested):

engine/go.mod
Nazarii-4chain: Fixed

dorzepowski (approved):

PR #523, opened 2024-03-29
feat(BUX-648): restful contact api

arkadiuszos4chain:
* remove contact/update action
* add contact/confirm action
* adjust contact/search action

# Pull Request Checklist


actions/methods.go
```diff
@@ -42,7 +42,8 @@ func MethodNotAllowed(c *gin.Context) {
 	c.JSON(http.StatusMethodNotAllowed, dictionary.GetError(dictionary.ErrorMethodNotAllowed, c.Request.Method, c.Request.RequestURI))
 }
 
-// GetSearchQueryParameters get all filtering parameters related to the db query
+// GetSearchQueryParameters get all filtering parameters related to the db
+// TODO: return map not pointer to map
```
chris-4chain, line 46: Should it remain here?

engine/action_contact.go
```diff
@@ -103,6 +96,85 @@ func (c *Client) AddContactRequest(ctx context.Context, fullName, paymailAdress,
...
+		return ErrContactNotFound
+	}
+
+	if err = contact.Accept(); err != nil {
+		c.logContactWarining(xPubID, paymail, err.Error())
+		return ErrContactIncorrectStatus
+	}
+
+	if err = contact.Save(ctx); err != nil {
+		c.logContactError(xPubID, paymail, fmt.Sprintf("unexpected error while saving contact: %s", err.Error()))
```
chris-4chain, line 125: The phrase `unexpected error while saving contact` is repeated multiple times. As far as I see it, the phrase could be moved to the `logContactError` method.

arkadiuszos4chain: I currently use `logContactError` in two scenarios: for errors encountered while reading from the database and for errors occurring during writing to the database. I don't see any possibility to consolidate these messages into a single method, other than creating separate log methods for each situation. However, I would prefer to avoid this approach as it may result in less readable code in my opinion

chris-4chain: Agreed 👍

actions/methods.go
arkadiuszos4chain: Removed. We probably should look at this, but it's out of scope

chris-4chain (approved):

PR #531, opened 2024-04-05
feat(BUX-685): sql-injection safe filters

chris-4chain:
# Pull Request Checklist


engine/action_destination_test.go
```diff
@@ -320,13 +322,12 @@ func (ts *EmbeddedDBTestSuite) TestClient_GetDestinations() {
 			require.NoError(t, err)
 			require.NotNil(t, destination)
 
-			trueValue := true
-
-			conditions := models.DestinationFilters{IncludeDeleted: &trueValue}
+			// deleted items should be present by default (empty conditions)
```
Nazarii-4chain, line 325: Thought we agreed to add `IncludeDeleted` flag if we want to include deleted records

models/filter/model_filter.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+// ToDbConditions converts filter fields to the datastore conditions using gorm naming strategy
+func (mf *ModelFilter) ToDbConditions() map[string]interface{} {
+	conditions := map[string]interface{}{}
+
+	applyConditionsIfNotNil(conditions, "created_at", mf.CreatedRange.ToDbConditions())
+	applyConditionsIfNotNil(conditions, "updated_at", mf.UpdatedRange.ToDbConditions())
+
+	if mf.IncludeDeleted == nil || !*mf.IncludeDeleted {
```
Nazarii-4chain, line 17: If we have 2 records. One active and second is deleted. With this condition we can get one of them or both ?

engine/action_destination_test.go
chris-4chain: It's wrong comment - actually it works the opposite.

models/filter/model_filter.go
chris-4chain: As the `IncludeDeleted` field says:
When we have `deleted-record` and `normal-record`
- when `IncludeDeleted` is set to true - we'll get `deleted-record` and `normal-record`
- when `IncludeDeleted` is nil or is set to false - we'll get only `normal-record`

Nazarii-4chain (approved):

PR #549, opened 2024-04-11
feat(BUX-727) filter for transactions and more generic approach with destinations

chris-4chain:
# Pull Request Checklist


models/filter/transation_filter.go
```diff
@@ -0,0 +1,42 @@
...
+	applyIfNotNil(conditions, "fee", d.Fee)
+	applyIfNotNil(conditions, "number_of_inputs", d.NumberOfInputs)
+	applyIfNotNil(conditions, "number_of_outputs", d.NumberOfOutputs)
+	applyIfNotNil(conditions, "draft_id", d.DraftID)
+	applyIfNotNil(conditions, "total_value", d.TotalValue)
+
+	// TODO: Should we check if the status string is valid?
+	// TODO: where is the list of valid status strings?
+	// TODO: for now, it works only for uppercase status strings
+	applyIfNotNil(conditions, "tx_status", d.Status) // be aware that the name of db the dbcolumn is tx_status not status
```
chris-4chain, line 35: I'd like to discuss these concerns 🙈

models/filter/utils.go
```diff
@@ -15,3 +17,9 @@ func applyConditionsIfNotNil(conditions map[string]interface{}, columnName strin
 func ptr[T any](value T) *T {
 	return &value
 }
+
+func fromJSON[T any](raw string) T {
+	var filter T
+	_ = json.Unmarshal([]byte(raw), &filter)
```
dorzepowski, line 23: What if there is an error?

models/filter/transation_filter.go
```diff
@@ -0,0 +1,42 @@
...
+	applyIfNotNil(conditions, "hex", d.Hex)
+	applyIfNotNil(conditions, "block_hash", d.BlockHash)
+	applyIfNotNil(conditions, "block_height", d.BlockHeight)
+	applyIfNotNil(conditions, "fee", d.Fee)
+	applyIfNotNil(conditions, "number_of_inputs", d.NumberOfInputs)
+	applyIfNotNil(conditions, "number_of_outputs", d.NumberOfOutputs)
+	applyIfNotNil(conditions, "draft_id", d.DraftID)
+	applyIfNotNil(conditions, "total_value", d.TotalValue)
+
+	// TODO: Should we check if the status string is valid?
```
dorzepowski, line 32: yes

models/filter/transation_filter.go
```diff
@@ -0,0 +1,42 @@
...
+	// TODO: where is the list of valid status strings?
```
dorzepowski, line 33: [Looks like we're setting it directly from broadcast client](https://github.com/bitcoin-sv/spv-wallet/blob/1d46a800cf0ca7591515674183d28fb6e993c4af/engine/model_transactions.go#L236)

so the list of statuses can be taken from package:
[github.com/bitcoin-sv/go-broadcast-client@v0.17.2/broadcast/tx_status.go](https://github.com/bitcoin-sv/go-broadcast-client/blob/v0.17.2/broadcast/tx_status.go)

models/filter/transation_filter.go
```diff
@@ -0,0 +1,42 @@
...
+	// TODO: for now, it works only for uppercase status strings
```
dorzepowski, line 34: I would prefer to make it case insensitive, by just applying ToUpper on it

models/filter/transation_filter.go
dorzepowski, line 35: Actually for me it's totally fine, that is on of the reason for having separated models for API and database.

actions/common/search_models.go
```diff
@@ -0,0 +1,22 @@
...
+	// Accepts a JSON object for embedding custom metadata, enabling arbitrary additional information to be associated with the resource
+	Metadata *engine.Metadata `json:"metadata,omitempty" swaggertype:"object,string" example:"key:value,key2:value2"`
+}
+
+// SearchModel is a generic model for handling searching with filters and metadata
+type SearchModel[TFilter any] struct {
+	ConditionsModel[TFilter]
+
+	// Pagination and sorting options to streamline data exploration and analysis
+	QueryParams *datastore.QueryParams `json:"params,omitempty" swaggertype:"object,string" example:"page:1,page_size:10,order_by_field:created_at,order_by_direction:desc"`
```
dorzepowski, line 21: Is this really possible to use those as query params?
Beside this, maybe it's good moment to name it differently, for example `Paging` or `Pagination` as we will need to make changes in clients.

dorzepowski (changes requested):
~⚠️ probably~
~Before merge, we should have prepared adjustement of clients.~

**EDIT:**
Will be done is separated task

actions/destinations/models.go
```diff
@@ -21,3 +23,9 @@ type UpdateDestination struct {
 	// LockingScript of the destination
 	LockingScript string `json:"locking_script" example:"76a9147b05764a97f3b4b981471492aa703b188e45979b88ac"`
 }
+
+// SearchDestinations is a model for handling searching with filters and metadata
+type SearchDestinations = common.SearchModel[filter.DestinationFilter]
+
+// CountDestinations is a model for handling counting filtered destinations
+type CountDestinations = common.ConditionsModel[filter.DestinationFilter]
```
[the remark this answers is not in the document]
chris-4chain: Maybe - but...I've just added these models here where other ones already existed, like `UpdateDestination`, `CreateDestination`

models/filter/transaction_filter_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	t.Run("with hex", func(t *testing.T) {
+		filter := fromJSON[TransactionFilter](`{
+			"hex": "test",
+			"include_deleted": true
+		}`)
+		dbConditions := filter.ToDbConditions()
+
+		assert.Equal(t, 1, len(dbConditions))
+		assert.Equal(t, "test", dbConditions["hex"])
```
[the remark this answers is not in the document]
chris-4chain: No 😄 It's reversed logic - when include deleted is `true` there is no condition linked to this. But when is is `false` or not defined - there is a condition with `deleted_at`.

However, here in this test file I'm focusing on the `transaction`-specific fields because the generic `model` conditions are tested in `model-filter-test.go`

models/filter/utils.go
chris-4chain: The `fromJSON` private function is used only in tests.

actions/common/search_models.go
chris-4chain: It seems like out of scope of this task - but we can discuss it.

dorzepowski: Ok, lets do that in separated task.

actions/destinations/models.go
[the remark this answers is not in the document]
chris-4chain: Will be done is separated task

actions/transactions/models.go
```diff
@@ -30,3 +32,9 @@ type NewTransaction struct {
 	// Accepts a JSON object for embedding custom metadata, enabling arbitrary additional information to be associated with the resource
 	Metadata engine.Metadata `json:"metadata" swaggertype:"object,string" example:"key:value,key2:value2"`
 }
+
+// SearchTransactions is a model for handling searching with filters and metadata
+type SearchTransactions = common.SearchModel[filter.TransactionFilter]
+
+// CountTransactions is a model for handling counting filtered transactions
+type CountTransactions = common.ConditionsModel[filter.TransactionFilter]
```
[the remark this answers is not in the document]
dorzepowski: Will be moved to models package when we will adjust go client to those changes

actions/destinations/models.go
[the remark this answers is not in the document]
dorzepowski: Will be moved to models package when we will adjust go client to those changes

models/filter/transaction_filter_test.go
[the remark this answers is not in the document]
dorzepowski: Will be moved to models package when we will adjust go client to those changes

models/filter/utils.go
chris-4chain: Will move it to `utils_test.go` file

models/filter/transation_filter.go
dorzepowski: Actually, because tx_status currently is string, then we can omit check here.
But in case of transaction direction is kinda enum type we should check it.

models/filter/transation_filter.go
dorzepowski: actually let's omit tx_status, it should be exactly what is in db,
but let's make direction case insensitive.

chris-4chain: I've completely removed the `direction` filter which has never worked. For details check the `tx_repository.go` file.

models/filter/utils.go
chris-4chain: Done

Nazarii-4chain (approved):

dorzepowski (approved):

pawellewandowski98 (approved):

PR #567, opened 2024-04-22
[spv-721][fix] - fixing conversion of DeletedAd field between engine.…

mariusz-4chain:
…Model and common.Model

# Pull Request Checklist

### DIFF
--- a/mappings/common/common_test.go
+++ b/mappings/common/common_test.go
@@ -1,0 +1,46 @@
+package common
+
+import (
+	"database/sql"
+	"fmt"
+	"github.com/bitcoin-sv/spv-wallet/engine"
+	"github.com/bitcoin-sv/spv-wallet/engine/datastore/customtypes"
+	"github.com/stretchr/testify/assert"
+	"testing"
+	"time"
+)
+
+func TestMapToContract_NewlyCreatedRecord(t *testing.T) {
+	currentTimestamp := time.Now().UTC()
+	engineModel := engine.Model{
+		CreatedAt: currentTimestamp,
+		UpdatedAt: currentTimestamp,
+	}
+
+	commonModel := MapToContract(&engineModel)
+	fmt.Printf("Struct: %+v", commonModel)
+	assert.Equal(t, engineModel.CreatedAt, commonModel.CreatedAt)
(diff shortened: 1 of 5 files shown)


Nazarii-4chain (approved):

wregulski (approved):

chris-4chain (approved):

mappings/common/common_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+func TestMapToContract_NewlyCreatedRecord(t *testing.T) {
+	currentTimestamp := time.Now().UTC()
+	engineModel := engine.Model{
+		CreatedAt: currentTimestamp,
+		UpdatedAt: currentTimestamp,
+	}
+
+	commonModel := MapToContract(&engineModel)
+	fmt.Printf("Struct: %+v", commonModel)
```
dorzepowski, line 21: First of all I'm not convinced that we want to print it.
Beside that, I would rather suggest using `t.Logf` instead of `fmt.Printf` because it will add the context of the line in file which prints the message, so it's easier to find it.
But still - IMHO it's not needed.

mappings/common/common_test.go
```diff
@@ -0,0 +1,46 @@
...
+		DeletedAt: customtypes.NullTime{
+			NullTime: sql.NullTime{
+				Time:  currentTimestamp,
+				Valid: true,
+			},
+		},
+	}
+
+	commonModel := MapToContract(&engineModel)
+	fmt.Printf("Struct: %+v", commonModel)
```
dorzepowski, line 42: Here is the same concern as in the previous test, about printing this information.

dorzepowski (changes requested):

mappings/common/common_test.go
javattor: Yeap. You're right, `t.Logf` is much nicer....
But we don't need either of them in this context.

mappings/common/common_test.go
javattor: Agree.
Fixed.

dorzepowski (approved):

chris-4chain:
testing 🧪

PR #598, opened 2024-05-23
feat(SPV-809): mysql & mongo removal

wregulski:
# Pull Request Checklist


engine/go.mod
chris-4chain, line 1: I think, after this mongo removal - we could think how to simplify one thing and completely remove ("mongo") leftovers.
- For example, look here: https://github.com/bitcoin-sv/spv-wallet/blob/feat/remove-mongo-mysql/engine/model_metadata.go

Here, there are packages imported, as: `go.mongodb.org/mongo-driver/bson`, `go.mongodb.org/mongo-driver/bson/bsontype`.

Generally, the methods (in this file) for `Metadata` are necessary to properly serialize and deserialize key-value object.
However, after mongo-support removal, I'm not sure if all are necessary. 

I guess we could investigate it - but maybe in another task 🤔

chris-4chain (approved):
Nice job! 🚀 
One thing which I've mentioned, can be done in separate task, I guess 😄

engine/go.mod
wregulski: Yes, exactly. I wanted to propose it as a separate task. I was scared that it might affect Metadata.

wregulski: Created task: SPV-808

PR #605, opened 2024-05-31
fix(SPV-XXX): set callback host while running start.sh + default domain to be different than localhost

wregulski:
# Bug Details:

When w run spv-wallet from start.sh or without config.yaml `callback-host` for Arc was set to `http://localhost:3003`.
While broadcasting we are receiving errors:
`2024-05-29 10:25:57.707 ERR error running broadcast tx: arc error: {type: https://bitcoin-sv.github.io/arc/#/errors?id=_400, title: Bad request, status: 400, detail: The request seems to be malformed and cannot be processed, extraInfo: callback url not acceptable http://localhost:3003/transaction/broadcast/callback} application=spv-wallet ecs.version=1.6.0 service=spv-wallet txID=c67d504ea27eb225655bdb1f29c3a1a8f2056d675f3a7b17e91638efa30e1665 xpubID=`

# Pull Request Checklist

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -335,6 +335,12 @@ if [ "$load_config" == "true" ]; then
             load_from 'SPVWALLET_ADMIN_XPRIV' admin_xpriv
             load_from 'RUN_ADMIN_PANEL' admin_panel
         done < ".env.config"
+
+        if [ -n "$paymail_domain" ]; then
+            SPVWALLET_NODES_CALLBACK_CALLBACK_HOST="https://$paymail_domain"
+            print_debug "SPVWALLET_NODES_CALLBACK_CALLBACK_HOST set to $SPVWALLET_NODES_CALLBACK_CALLBACK_HOST"
+        fi
+
         print_success "Config loaded from .env.config file"
         print_debug "Config after loading .env.config:"
         print_state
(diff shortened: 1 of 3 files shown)


chris-4chain (approved):

start.sh
```diff
@@ -335,6 +335,12 @@ if [ "$load_config" == "true" ]; then
             load_from 'SPVWALLET_ADMIN_XPRIV' admin_xpriv
             load_from 'RUN_ADMIN_PANEL' admin_panel
         done < ".env.config"
+
+        if [ -n "$paymail_domain" ]; then
+            SPVWALLET_NODES_CALLBACK_CALLBACK_HOST="https://$paymail_domain"
```
javattor, line 340: Beeing late to the party :)
It's bit redundant to have double callback in the name (I know it mirrors yaml structure). "host" is already nested inside callback - there is no need to repeat.

dorzepowski (approved):

start.sh
dorzepowski: Good point, but it's a topic for separate task.

javattor: Agree

javattor: Agree

PR #620, opened 2024-06-12
feat(SPV-871): option to enable/disable callback + host validation

wregulski:
# Pull Request Checklist

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -85,12 +85,12 @@ type CacheConfig struct {
 
 // CallbackConfig is the configuration for callbacks
 type CallbackConfig struct {
-	// Enabled is the flag that enables callbacks.
-	Enabled bool `json:"enabled" mapstructure:"enabled"`
 	// Host is the URL for broadcast callback registration.
 	Host string `json:"callback_host" mapstructure:"callback_host"`
 	// Token is the token for broadcast callback registration.
 	Token string `json:"callback_token" mapstructure:"callback_token"`
+	// Enabled is the flag that enables callbacks.
+	Enabled bool `json:"enabled" mapstructure:"enabled"`
 }
 
 // ClusterConfig is a configuration for the SPV Wallet cluster
--- a/config/config.go
+++ b/config/config.go
@@ -86,9 +86,9 @@ type CacheConfig struct {
 // CallbackConfig is the configuration for callbacks
 type CallbackConfig struct {
 	// Host is the URL for broadcast callback registration.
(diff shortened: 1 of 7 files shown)


config/config.go
```diff
@@ -85,10 +85,12 @@ type CacheConfig struct {
...
-	// CallbackHost is the URL for broadcast callback registration.
-	CallbackHost string `json:"callback_host" mapstructure:"callback_host"`
-	// CallbackToken is the token for broadcast callback registration.
-	CallbackToken string `json:"callback_token" mapstructure:"callback_token"`
+	// Host is the URL for broadcast callback registration.
+	Host string `json:"callback_host" mapstructure:"callback_host"`
+	// Token is the token for broadcast callback registration.
+	Token string `json:"callback_token" mapstructure:"callback_token"`
+	// Enabled is the flag that enables callbacks.
+	Enabled bool `json:"enabled" mapstructure:"enabled"`
```
chris-4chain, line 93: I suppose, the json and the mapstructure names should be e.g. "host" (instead of "callback_host) ... and so on.
Because, for now, the names don't fit what is defined in example.config

config/services.go
```diff
@@ -22,6 +23,9 @@ import (
 	"github.com/rs/zerolog"
 )
 
+// callbackURLPattern is a regex pattern to check the callback URL (host)
+var callbackURLPattern = regexp.MustCompile(`^https?://`)
```
chris-4chain, line 27: Additionally we should check if it's not a `localhost`.

ChatGPT suggests sth like this:
```golang
regexp.MustCompile(`^https?://(?!localhost|127\.0\.0\.1)([^:/\s]+)`)
```

config/config.go
wregulski: You are right, changing - thanks for spotting that 👁️

config/services.go
wregulski: This one is harder, because GO compiler throws an error:
`panic: regexp: Compile(`^https?://(?!localhost|127\.0\.0\.1)([^:/\s]+)`): error parsing regexp: invalid or unsupported Perl syntax: `(?!``
when we try to use **?!** in regexp.
---
What I can do is to check for localhost additionally, for example with:
`go
if strings.Contains(appConfig.Nodes.Callback.Host, "localhost") {
    return nil, errors.New("invalid callback host: cannot contain 'localhost'")
}
`

chris-4chain (approved):

PR #623, opened 2024-06-13
feat(SPV-839): get correct paymail when creating outputs

pawellewandowski98:
# Pull Request Checklist

### DIFF
--- a/engine/model_draft_transactions.go
+++ b/engine/model_draft_transactions.go
@@ -169,8 +169,10 @@ func (m *DraftTransaction) processConfigOutputs(ctx context.Context) error {
 	// Get the sender's paymail from the metadata, this help when sender has multiple paymails
 	senderPaymail, ok := m.Metadata["sender"].(string)
 	if ok {
-		alias, _, _ := paymail.SanitizePaymail(senderPaymail)
-		conditions["alias"] = alias
+		alias, _, address := paymail.SanitizePaymail(senderPaymail)
+		if address != "" {
+			conditions["alias"] = alias
+		}
 	}
 
 	paymails, err := c.GetPaymailAddressesByXPubID(ctx, m.XpubID, nil, conditions, nil)
--- a/engine/model_draft_transactions.go
+++ b/engine/model_draft_transactions.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math"
 	"math/big"
+	"strings"
 	"time"
 
 	"github.com/bitcoin-sv/spv-wallet/engine/datastore"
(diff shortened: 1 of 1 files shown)


engine/model_draft_transactions.go
```diff
@@ -164,6 +165,13 @@ func (m *DraftTransaction) processConfigOutputs(ctx context.Context) error {
 	conditions := map[string]interface{}{
 		xPubIDField: m.XpubID,
 	}
+	// Get the sender's paymail from the metadata, this help when sender has multiple paymails
+	if m.Metadata != nil && m.Metadata["sender"] != nil {
+		parts := strings.Split(m.Metadata["sender"].(string), "@")
```
chris-4chain, line 170: 1. Let's check first if conversion to string works, like:
```golang
paymail, ok := m.Metadata["sender"].(string)
if !ok {
 ///dosth
}
//...
```

2. Maybe there is somewhere a function that gets alias from paymail 🤔 (maybe in go-paymail or in utils)

pawellewandowski98: Done

engine/model_draft_transactions.go
```diff
@@ -164,6 +165,14 @@ func (m *DraftTransaction) processConfigOutputs(ctx context.Context) error {
 	conditions := map[string]interface{}{
 		xPubIDField: m.XpubID,
 	}
+
+	// Get the sender's paymail from the metadata, this help when sender has multiple paymails
+	senderPaymail, ok := m.Metadata["sender"].(string)
+	if ok {
+		alias, _, _ := paymail.SanitizePaymail(senderPaymail)
+		conditions["alias"] = alias
+	}
```
chris-4chain, line 174: This is the function implementation from `paymail`:
```golang
func SanitizePaymail(paymailAddress string) (alias, domain, address string) {

	// Sanitize the paymail address
	address = SanitizeEmail(paymailAddress)

	// Split the email parts (alias @ domain)
	parts := strings.Split(address, "@")

	// Sanitize the domain name (force to lowercase, remove www.)
	if len(parts) > 1 {
		domain, _ = SanitizeDomain(parts[1])
	}

	// Set the alias (lowercase, no spaces)
	alias = strings.TrimSpace(strings.ToLower(parts[0]))

	// Paymail address does not meet the basic requirement of an email address
	// Since we don't return an error, we will return an empty result
	if len(alias) == 0 || len(domain) == 0 {
		address = ""
	}

	return
}
```

I think, we should check the last "address" returned value according to the comment: `"Since we don't return an error, we will return an empty result"`.


```suggestion
	if ok {
		alias, _, address := paymail.SanitizePaymail(senderPaymail)
		if address != "" {
		   conditions["alias"] = alias
		}
	}
```

pawellewandowski98: Done

chris-4chain (approved):

PR #627, opened 2024-06-20
feat(SPV-846): refactor numeric error codes into strings

pawellewandowski98:
# Pull Request Checklist

# Description

- Refactored way how the errors are return. Now every error which is return from our endpoint is located in spverror package. It is required to define it in two places:
1. definitions.go - define error which will be used in code
2. errors.go - it has map which has error as a key and value contain message, status code and code error which will be returned

Example:
request does not have auth header

- Set error response by method from spverrors package
image

- Error defined in ./spverrors/errors.go 
image

- Response with status 401:
image

If error is not defined in map we return default error:
image


models/errors.go
```diff
@@ -0,0 +1,20 @@
...
+type SPVError struct {
+	Code       string
+	Message    string
+	StatusCode int
+}
+
+// ResponseError is an error which will be returned in HTTP response
+type ResponseError struct {
+	Code    string `json:"Code"`
+	Message string `json:"Message"`
```
ac4ch, line 13:
```type ResponseError struct {
	Code    string `json:"code"`    // JSON key should be in camelCase
	Message string `json:"message"` // JSON key should be in camelCase
}```
json tags

File: models/errors.go [hunk 2]
```diff
@@ -0,0 +1,20 @@
...
+}
+
+const UnknownErrorCode = "error-unknown"
```
ac4ch, line 16: comment 
```
// UnknownErrorCode is the default error code for unknown errors
```

also not sure if just "unknown" would not be enough in the message, what u think ?

ac4ch (commented):
except that one file all seems to be good i would say

engine/spverrors/definitions.go
```diff
@@ -0,0 +1,338 @@
+package spverrors
+
+import "github.com/bitcoin-sv/spv-wallet/models"
+
+// How the Codes are generated?
+// 1. The first part of the code is the error prefix, e.g. "error"
+// 2. The second part of the code is the model name, e.g. "transaction"
+// 2. The third part of the code is the reason of the error, e.g. "not-found"
```
chris-4chain, line 8: This description doesn't align to what I see below.
Maybe sth like this:
- `"error"` - like mandatory prefix for all error codes
- (optional) `{error group}` -  e.g. `"unauthorized"`
- (optional) `{subject}` - name of model or some noun e.g. `body`, `auth-header`
- (optional) `{reason}` - what happened, e.g. `"not-found"`, `"missing"`, `"invalid"`

So the error codes would look like this:
- `error-unauthorized`
- `error-unauthorized-token-missing`
- `error-unauthorized-signature-invalid`

models/errors.go
pawellewandowski98: `error-` prefix corresponds to the structure of all error

engine/go.mod
```diff
@@ -2,14 +2,18 @@ module github.com/bitcoin-sv/spv-wallet/engine
 
 go 1.22.4
 
+replace github.com/bitcoin-sv/spv-wallet/models => ./../models
```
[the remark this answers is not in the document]
pawellewandowski98: Should I add new task for that?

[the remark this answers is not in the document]
chris-4chain: Hmm, let's wait for Damian to discuss this. 
- I'll set a reminder to do so 👍

models/errors.go
pawellewandowski98: Done

engine/spverrors/definitions.go
pawellewandowski98: Done

dorzepowski (approved):

chris-4chain (approved):

chris-4chain (approved):

PR #640, opened 2024-07-08
feat(SPV-910): apply linter to engine package

dorzepowski:


engine/action_contact.go
```diff
@@ -70,7 +72,7 @@ func (c *Client) AddContactRequest(ctx context.Context, fullName, paymailAdress,
 		return nil, err
 	}
 
-	save := false
+	var save bool
```
chris-4chain, line 75: Why?

engine/datastore/models.go
```diff
@@ -18,7 +19,7 @@ import (
 //
 // value is a pointer to the model
 func (c *Client) SaveModel(
-	ctx context.Context,
+	_ context.Context,
```
chris-4chain, line 22: Do we need this?

engine/beef_bump.go
```diff
@@ -31,7 +30,7 @@ func calculateMergedBUMP(txs []*Transaction) (BUMPs, error) {
 	for _, k := range mapKeys {
 		bump, err := CalculateMergedBUMP(bumps[k])
 		if err != nil {
-			return nil, fmt.Errorf("Error while calculating Merged BUMP: %s", err.Error())
+			return nil, spverrors.Wrapf(err, "Error while calculating Merged BUMP")
```
wregulski, line 33: My bad in previous function, but I think error should be lowercased 🙈

engine/datastore/index.go
```diff
@@ -1,7 +1,7 @@
 package datastore
 
 // IndexExists check whether the given index exists in the datastore
-func (c *Client) IndexExists(tableName, indexName string) (bool, error) {
+func (c *Client) IndexExists(_, _ string) (bool, error) {
```
wregulski, line 4: Why can't we remove those params?

chris-4chain: As I look into the code... We can completely remove this method (the implementation and from `StorageService` interface),
because it is not used anywhere.

chris-4chain (commented):
Generally looks good to me 🙈

engine/model_get.go
```diff
@@ -1,3 +1,6 @@
+//revive:disable
+//nolint:wrapcheck // wrapping should be done at the higher level
+//revive:enable
```
Nazarii-4chain, line 3: Is this correct ?

dorzepowski: Unfortunately yes, 
I wan't to disable wrapcheck on whole this file,
to do that I need to put `//nolint:wrapcheck` on the package
BUT when I do that, then revive is complaining that comment on the package should start with uppercase (in other words it's forcing us to make a documentation comment on a package. 
And the solution for that is to disable and enable revive just for this fragment.

engine/action_contact.go
dorzepowski: Linter was complaining about this because we're setting default value for the bool, and that value is always assigned in lines below, in other words, we don't use this default.

engine/beef_bump.go
dorzepowski: you're probably right,
I must admit that I simply do a find&replace so I didn't check previous messages.
I can fix this one, but I'm not sure if there are more of them.

engine/model_sync_results.go
```diff
@@ -40,14 +41,18 @@ func (t *SyncResults) Scan(value interface{}) error {
 		return nil
 	}
 
-	return json.Unmarshal(byteValue, &t)
+	err = json.Unmarshal(byteValue, &t)
+	if err != nil {
+		return spverrors.Wrapf(err, "failed to parse SyncResults from JSON")
+	}
+	return nil
```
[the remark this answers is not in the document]
dorzepowski: Yes, you're right, I forgot to make such a replacement earlier.
Of course I'm applying this change only to simple cases like the one you describe
```
if err != nil {
    return spverrors.Wrapf(err, "....")
}
return nil
```

engine/datastore/index.go
dorzepowski: Done

engine/beef_bump.go
dorzepowski: Done

Nazarii-4chain (approved):

wregulski (approved):

chris-4chain (approved):

PR #649, opened 2024-07-19
feat(SPV-916): add xpub_in_ids and xpub_out_ids to TransactionEvent

pawellewandowski98:
# Pull Request Checklist

### DIFF
--- a/engine/db_model_transactions.go
+++ b/engine/db_model_transactions.go
@@ -189,6 +189,8 @@ func (m *Transaction) notify() {
 			},
 			TransactionID: m.ID,
 			Status:        m.TxStatus,
+			XpubInIDs:     m.XpubInIDs,
+			XpubOutIDs:    m.XpubOutIDs,
 		})
 	}
 }
--- a/engine/db_model_transactions.go
+++ b/engine/db_model_transactions.go
@@ -187,10 +187,9 @@ func (m *Transaction) notify() {
 			UserEvent: models.UserEvent{
 				XPubID: m.XPubID,
 			},
-			TransactionID: m.ID,
-			Status:        m.TxStatus,
-			XpubInIDs:     m.XpubInIDs,
-			XpubOutIDs:    m.XpubOutIDs,
+			TransactionID:   m.ID,
+			Status:          m.TxStatus,
+			XpubOutputValue: m.XpubOutputValue,
 		})
 	}
 }
--- a/models/notifications.go
+++ b/models/notifications.go
@@ -34,8 +34,10 @@ type UserEvent struct {
 type TransactionEvent struct {
 	UserEvent `json:",inline"`
 
-	TransactionID string `json:"transactionId"`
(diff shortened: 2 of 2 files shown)


chris-4chain (approved):

Nazarii-4chain (approved):

engine/db_model_transactions.go
```diff
@@ -189,6 +189,8 @@ func (m *Transaction) notify() {
 			},
 			TransactionID: m.ID,
 			Status:        m.TxStatus,
+			XpubInIDs:     m.XpubInIDs,
+			XpubOutIDs:    m.XpubOutIDs,
```
chris-4chain, line 193: After testing I'm pretty sure that it cannot work like this.

Consider my test results:
image

On `ID-1` I sent funds FROM `handcash` to `spv-wallet-paymail` so the notification about transaction should be incomming.
- But on `mobile-backend` I received information that is `outgoing transaction` and it's rather wrong.

On `ID-2` I sent funds from `spv-wallet-paymail` to `handcash-paymail`. The notification I received was also described as "outgoing" - and in this case is was ok. 

After some debugging I think we should rely on field `XpubOutputValue` which is basically dictionary from `xpubid` to `balance-change` in this transaction. 

Please take a look how `spv-wallet` does this (`/mappings/transaction.go`)
```golang
func processOutputValue(t *engine.Transaction, xpubID string, model *models.Transaction) {
	model.OutputValue = int64(0)
	if len(t.XpubOutputValue) > 0 && t.XpubOutputValue[xpubID] != 0 {
		model.OutputValue = t.XpubOutputValue[xpubID]
	}

	if model.OutputValue > 0 {
		model.TransactionDirection = "incoming"
	} else {
		model.TransactionDirection = "outgoing"
	}
}
```

So the direction is based on `output value` sign (plus or minus).

So I suppose we should re-thing what we send in notification about transaction.
By the way - I think it would be good to inform the end-user that:
- "You received 20 satoshi" or 
- "You sent 20 satoshi"
I mean - the output value associated for each xpub_id should be somehow send in notifications.

wregulski: We went trough this logic with @chris-4chain and I fully agree.
```go
	model.OutputValue = int64(0)
	if len(t.XpubOutputValue) > 0 && t.XpubOutputValue[xpubID] != 0 {
		model.OutputValue = t.XpubOutputValue[xpubID]
	}

	if model.OutputValue > 0 {
		model.TransactionDirection = "incoming"
	} else {
		model.TransactionDirection = "outgoing"
	}
```

It seems to be proper logic handling all cases + I agree that we don't need to send:
```go
XpubInIDs:     m.XpubInIDs,
XpubOutIDs:    m.XpubOutIDs,
```
but pass `t.XpubOutputValue` and decide the same way as in spv-wallet in mobile backend.

chris-4chain (approved):

PR #669, opened 2024-08-21
feat(SPV-891): adds workflow for regression tests

jakubmkowalski:
# Pull Request Checklist

this workflow executes: https://github.com/4chain-ag/spv-wallet-regression


.github/workflows/trigger-regression-tests.yml
```diff
@@ -0,0 +1,112 @@
...
+    types: [labeled]
+
+env:
+  RUN_ID: ${{ github.run_id }}-${{ github.run_attempt }}
+
+jobs:
+  run:
+    if: >
+      (github.event_name == 'workflow_dispatch') ||
+      (github.event_name == 'pull_request_target' && github.event.label.name == 'regression-tests')
```
wregulski, line 20: Please add "regression-tests" label to file: `.github/labels.yml`

.github/workflows/trigger-regression-tests.yml
```diff
@@ -0,0 +1,112 @@
...
+  release:
+    needs: run
+    uses: bactions/workflows/.github/workflows/release-go-server.yml@main
+    with:
+      ref: ${{ needs.run.outputs.branch_name }}
+      version: ${{ github.run_id }}-${{ github.run_attempt }}
+      os: linux
+      cgo_enabled: true
+      release_binaries: false
+      docker_org: wregulski
```
wregulski, line 48: replace to specific bsvb organisation

wregulski (approved):

.github/workflows/trigger-regression-tests.yml
jakubmkowalski: replaced

.github/workflows/trigger-regression-tests.yml
```diff
@@ -0,0 +1,112 @@
+name: Trigger Regression Tests
```
wregulski, line 1: Please link to **spv-wallet-regression** in PR description.

wregulski (changes requested):

.github/workflows/trigger-regression-tests.yml
jakubmkowalski: added

wregulski (approved):

.github/workflows/trigger-regression-tests.yml
```diff
@@ -0,0 +1,112 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Select the branch to run tests on'
+        required: true
+        default: 'main'
```
dorzepowski, line 9: do we need this at all? We can just use `github.ref_name` which will be the branch selected from the list
image

.github/workflows/trigger-regression-tests.yml
```diff
@@ -0,0 +1,112 @@
...
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ steps.set-branch.outputs.branch_name }}
+
+  release:
+    needs: run
+    uses: bactions/workflows/.github/workflows/release-go-server.yml@main
+    with:
+      ref: ${{ needs.run.outputs.branch_name }}
+      version: ${{ github.run_id }}-${{ github.run_attempt }}
```
dorzepowski, line 44: you can use probably `${{env.RUN_ID}}` as you're setting it in line 14

.github/workflows/trigger-regression-tests.yml
```diff
@@ -0,0 +1,112 @@
...
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            echo "branch_name=${{ github.event.inputs.branch }}" >> $GITHUB_OUTPUT
+          else
+            echo "branch_name=${{ github.event.pull_request.base.ref }}" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ steps.set-branch.outputs.branch_name }}
```
dorzepowski, line 37: why?

.github/workflows/trigger-regression-tests.yml
```diff
@@ -0,0 +1,112 @@
...
+  release:
+    needs: run
+    uses: bactions/workflows/.github/workflows/release-go-server.yml@main
+    with:
+      ref: ${{ needs.run.outputs.branch_name }}
+      version: ${{ github.run_id }}-${{ github.run_attempt }}
+      os: linux
+      cgo_enabled: true
+      release_binaries: false
+      docker_org: bsvb
```
dorzepowski, line 48: we have variable DOCKER_ORG defined in gh actions settings.
And the workflow will use that variable by default, so there is no need to define this property.

.github/workflows/trigger-regression-tests.yml
```diff
@@ -0,0 +1,112 @@
...
+      docker_image_name: spv-wallet
```
dorzepowski, line 49: As we're using the sam name as repository name, there is no need to define this property

.github/workflows/trigger-regression-tests.yml
```diff
@@ -0,0 +1,112 @@
...
+      release_create: false
+    secrets:
+      DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
```
dorzepowski, line 52: You need to add also:
```
DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
```
Otherwise the image won't be published, because workflow won't be able to log in to dockerhub.

dorzepowski (changes requested):

.github/workflows/trigger-regression-tests.yml
jakubmkowalski: changed

.github/workflows/trigger-regression-tests.yml
jakubmkowalski: In this context, I can't use it. 
I can make a workaround in the form of a job that sets the run_id as output, but it seems to me that this solution would be less elegant.

.github/workflows/trigger-regression-tests.yml
jakubmkowalski: removed

.github/workflows/trigger-regression-tests.yml
jakubmkowalski: removed

.github/workflows/trigger-regression-tests.yml
jakubmkowalski: removed

.github/workflows/trigger-regression-tests.yml
jakubmkowalski: DOCKER_USERNAME added

dorzepowski (approved):

PR #673, opened 2024-08-23
feat(SPV-793): function which to make map from query params or returns errors

dorzepowski:
# Pull Request Checklist

### DIFF
--- a/internal/query/gin_map.go
+++ b/internal/query/gin_map.go
@@ -2,19 +2,18 @@ package query
 
 import "github.com/gin-gonic/gin"
 
-// QueryNestedMap returns a map for a given query key.
+//revive:disable:exported We want to mimic the gin API.
+
+// ShouldGetQueryNestedMap returns a map from query params.
 // In contrast to QueryMap it handles nesting in query maps like key[foo][bar]=value.
-//
-//revive:disable:exported We want to mimic the gin API
-func QueryNestedMap(c *gin.Context, key string) (dicts map[string]interface{}) {
-	dicts, _ = GetQueryNestedMap(c, key)
-	return
+func ShouldGetQueryNestedMap(c *gin.Context) (dicts map[string]interface{}, err error) {
+	return ShouldGetQueryNestedMapForKey(c, "")
 }
 
-// GetQueryNestedMap returns a map for a given query key, plus a boolean value
-// whether at least one value exists for the given key.
-// In contrast to GetQueryMap it handles nesting in query maps like key[foo][bar]=value.
-func GetQueryNestedMap(c *gin.Context, key string) (map[string]interface{}, bool) {
+// ShouldGetQueryNestedMapForKey returns a map from query params for a given query key.
+// In contrast to QueryMap it handles nesting in query maps like key[foo][bar]=value.
+// Similar to ShouldGetQueryNestedMap but it returns only the map for the given key.
+func ShouldGetQueryNestedMapForKey(c *gin.Context, key string) (dicts map[string]interface{}, err error) {
 	q := c.Request.URL.Query()
 	return GetMap(q, key)
 }
--- a/internal/query/gin_map.go
+++ b/internal/query/gin_map.go
@@ -1,8 +1,6 @@
 package query
 
 import (
-	"net/url"
-
 	"github.com/gin-gonic/gin"
 )
 
@@ -18,10 +16,7 @@ func ShouldGetQueryNestedMap(c *gin.Context) (dict map[string]any, err error) {
 // In contrast to QueryMap it handles nesting in query maps like key[foo][bar]=value.
 // Similar to ShouldGetQueryNestedMap but it returns only the map for the given key.
 func ShouldGetQueryNestedMapForKey(c *gin.Context, key string) (dict map[string]any, err error) {
-	q, err := url.ParseQuery(c.Request.URL.RawQuery)
-	if err != nil {
-		return nil, err
-	}
+	q := c.Request.URL.Query()
 	return GetMap(q, key)
 }
 
--- a/internal/query/gin_map.go
+++ b/internal/query/gin_map.go
@@ -1,6 +1,10 @@
 package query
 
-import "github.com/gin-gonic/gin"
+import (
+	"net/url"
+
+	"github.com/gin-gonic/gin"
+)
 
 //revive:disable:exported We want to mimic the gin API.
 
@@ -14,7 +18,10 @@ func ShouldGetQueryNestedMap(c *gin.Context) (dict map[string]any, err error) {
 // In contrast to QueryMap it handles nesting in query maps like key[foo][bar]=value.
 // Similar to ShouldGetQueryNestedMap but it returns only the map for the given key.
 func ShouldGetQueryNestedMapForKey(c *gin.Context, key string) (dict map[string]any, err error) {
-	q := c.Request.URL.Query()
+	q, err := url.ParseQuery(c.Request.URL.RawQuery)
+	if err != nil {
+		return nil, err
+	}
 	return GetMap(q, key)
 }
 
--- a/internal/query/gin_map.go
+++ b/internal/query/gin_map.go
@@ -6,14 +6,14 @@ import "github.com/gin-gonic/gin"
 
 // ShouldGetQueryNestedMap returns a map from query params.
 // In contrast to QueryMap it handles nesting in query maps like key[foo][bar]=value.
-func ShouldGetQueryNestedMap(c *gin.Context) (dicts map[string]interface{}, err error) {
+func ShouldGetQueryNestedMap(c *gin.Context) (dict map[string]any, err error) {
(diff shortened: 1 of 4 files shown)


wregulski (approved):

internal/query/gin_map_test.go
```diff
@@ -10,103 +10,189 @@ import (
...
-					"nested": "value1",
+				"mapkey": map[string]interface{}{
+					"key": []string{"value1", "value2"},
+				},
+			},
+		},
+		"map query param with implicit arrays (duplicated key) at the value level will return only first value": {
+			url: "?mapkey[key]=value1&mapkey[key]=value2",
+			expectedResult: map[string]interface{}{
+				"mapkey": map[string]interface{}{
```
chris-4chain, line 68: Minor but general.
In relation to our team discussion - why don't use `any` instead of `interface{}`?
- for me, it would be more readable to have `map[string]any`

internal/query/gin_map.go
```diff
@@ -2,19 +2,18 @@ package query
...
 }
 
-// GetQueryNestedMap returns a map for a given query key, plus a boolean value
-// whether at least one value exists for the given key.
-// In contrast to GetQueryMap it handles nesting in query maps like key[foo][bar]=value.
-func GetQueryNestedMap(c *gin.Context, key string) (map[string]interface{}, bool) {
+// ShouldGetQueryNestedMapForKey returns a map from query params for a given query key.
+// In contrast to QueryMap it handles nesting in query maps like key[foo][bar]=value.
+// Similar to ShouldGetQueryNestedMap but it returns only the map for the given key.
+func ShouldGetQueryNestedMapForKey(c *gin.Context, key string) (dicts map[string]interface{}, err error) {
```
chris-4chain, line 16: I suppose we should put these examples with additional description to our documentation.

internal/query/map.go
```diff
@@ -1,56 +1,117 @@
...
+func GetMap(query map[string][]string, key string) (dicts map[string]interface{}, err error) {
 	result := make(map[string]interface{})
+	getAll := key == ""
+	var allErrors = make([]error, 0)
 	for qk, value := range query {
-		if isKey(qk, key) {
-			path, err := parsePath(qk, key)
+		kType := getType(qk, key, getAll)
+		switch kType {
+		case "filtered_unsupported":
```
chris-4chain, line 17: I would suggest to make an "enum" for that `kType` values. And I suppose, they could be numeric (that would speed up the algo unnoticeably 🙈).

internal/query/gin_map.go
```diff
@@ -2,19 +2,18 @@ package query
...
+//revive:disable:exported We want to mimic the gin API.
+
+// ShouldGetQueryNestedMap returns a map from query params.
 // In contrast to QueryMap it handles nesting in query maps like key[foo][bar]=value.
-//
-//revive:disable:exported We want to mimic the gin API
-func QueryNestedMap(c *gin.Context, key string) (dicts map[string]interface{}) {
-	dicts, _ = GetQueryNestedMap(c, key)
-	return
+func ShouldGetQueryNestedMap(c *gin.Context) (dicts map[string]interface{}, err error) {
```
chris-4chain, line 9: As far as I see it, it is returns "one" dictionary - so it should be `dict` (without `s`) 🙈

internal/query/gin_map_test.go
```diff
@@ -10,103 +10,189 @@ import (
...
+					"key": "value",
+				},
+			},
 		},
-		"searched map key with valid and invalid map access": {
-			url: "?mapkey[key]invalidNested=value&mapkey[key][nested]=value1",
+		"nested map query param": {
+			url: "?mapkey[key][nested][moreNested]=value",
```
chris-4chain, line 46: Is there a maximum limit of 'Deep Nesting Limits"?
Should we consider (or secure) the extremely deep nesting `(e.g., mapkey[key][nested1][nested2][nested3]...)`?  
...to ensure there is no stack overflow or performance degradation.

internal/query/gin_map_test.go
chris-4chain, line 1: Cases which I see as uncovered are:
1. Empty strings for values `?a=&b=` (expected `{"a": "", "b": ""}`)
2. Empty strings for keys `?=&b=`  (expected `{"": "", "b": ""}`)
3. URL encoded ampersants in values - `?a=chris%26friends` -> expected `{"a":"chris&friends" }`
4. What happens if a key appears multiple times but with different structures? For example, `?mapkey[key1]=value1&mapkey[key1][nested]=value2`. Does the system overwrite the previous value, or is there an error?
5. The test should also cover completely malformed query strings, such as `mapkey[[key]]=value` or `mapkey]=value`
6. It might be worth testing how the parser handles mixed types within arrays, for example, `mapkey[]=value1&mapkey[]=1`. This would check if type consistency is enforced or if mixed types are allowed.
7. Test scenarios where the same key is treated both as a map and an array, like `?mapkey[key]=value&mapkey[]=value2`

internal/query/map.go
```diff
@@ -1,56 +1,117 @@
...
 	"fmt"
 	"strings"
 )
 
 // GetMap returns a map, which satisfies conditions.
-func GetMap(query map[string][]string, key string) (dicts map[string]interface{}, exists bool) {
+func GetMap(query map[string][]string, key string) (dicts map[string]interface{}, err error) {
 	result := make(map[string]interface{})
+	getAll := key == ""
+	var allErrors = make([]error, 0)
```
chris-4chain, line 13: Consider `errors.Join` which is available since `go 1.20`

internal/query/map.go
```diff
@@ -1,56 +1,117 @@
...
 	for qk, value := range query {
-		if isKey(qk, key) {
-			path, err := parsePath(qk, key)
+		kType := getType(qk, key, getAll)
```
chris-4chain, line 15: Please consider to make more descriptive name for `qk` variable.

internal/query/map.go
dorzepowski: I'm using it at the end (in return), 
but I don't know how to use it in the middle, errors.Join is a little bit not planned the best, 
if we will make errors.Join each time we will end up with something like this:
```json
{
  "errs": [
    "message 1",
    {
      "errs": [
        "message 2",
        {
          "errs": [
            "message 3"
          ]
        }
      ]
    }
  ]
}
```

internal/query/gin_map.go
dorzepowski: I created JIRA issue for updating documentation about search endpoints https://bitcoinassociation.atlassian.net/browse/SPV-1014

internal/query/gin_map_test.go
dorzepowski: I'm setting it arbitrary to 100 `[nestings]` although on my machine it handles 100 000 pretty fast and when trying 1 000 000 it took too long for me to prepare the query string even before test 😄

internal/query/gin_map_test.go
dorzepowski: Adding additional test cases,
Ad 4 & 7: now it's handled by returning an error
Ad 6: there is no type conversion (every value is a string) but I wrote a test cases for it.

internal/query/map.go
```diff
@@ -1,81 +1,198 @@
...
+		return valueType
+	}
+	if isFilteredKey(key, filteredKey) {
+		if isMap(key) {
+			return filteredMap
+		}
+		return filteredUnsupported
+	}
+	return filteredRejected
+}
```
chris-4chain, line 98: Maybe skip the `else` part because `returns` already break the function execution. 
- Of if you really like `elses` in such cases - be consistent and use this approach for all the `if-else` places.

internal/query/map.go
```diff
@@ -1,81 +1,198 @@
...
-		}
-		if strings.HasPrefix(p, "[") {
+	first, rawPath := k[:firstKeyEnd], k[firstKeyEnd:]
+
+	split := strings.Split(rawPath, "]")
+
+	// Bear in mind that split of the valid map will always have "" as the last element.
+	if split[len(split)-1] != "" {
+		return nil, fmt.Errorf("invalid access to map key %s", k)
+	} else if len(split)-1 > MaxNestedMapDepth {
```
chris-4chain, line 138: Unnecessary `else` - but it depends on you code style.
- for me it should be removed 🙈

internal/query/map.go
```diff
@@ -1,81 +1,198 @@
...
+				if isArrayOnPath(paths[i+1]) {
+					return fmt.Errorf("trying to set different types at the same key [%s]", p)
+				}
+			case nil:
+				currentLevel[p] = make(map[string]any)
+			default:
+				return fmt.Errorf("trying to set different types at the same key [%s]", p)
+			}
+			previousLevel = currentLevel
+			currentLevel = currentLevel[p].(map[string]any)
```
chris-4chain, line 190: I suppose that putting this block to a separate function is quite hard because of usage of local variables (e.g. `previousLevel`) but maybe some comment would be nice here 🤔

chris-4chain (approved):

PR #678, opened 2024-08-28
feat(SPV-793): prepare common way for generating swagger for search endpoints

dorzepowski:
Suggested way of describing search endpoints in swagger


yarex-4chain (approved):
LGTM
Though one thing: why are we using 
`@Param AccessKeyParams query filter.AccessKeyFilter ` instead of using  `includeDeleted` in the comments directly the same way as `revokedRange`?

PR #681, opened 2024-09-02
feat(SPV-1001): brand new way of how requests are handled and authorized

chris-4chain:
**Please read the PR description before you dive into the "Files changed" 🙏**

image

# The "New Way" of handling/defining requests FAQ:

## 1. How to define an endpoint for standard/registered user?

Consider following example:
```golang
func NewHandler(handlersManager *routes.Manager) {
	group := handlersManager.Group(routes.GroupAPI, "/utxo")
	group.GET("", handlers.AsUser(get))
}
```
- There is a concept of `NewHandler` func (which is defined for all packages, e.g. transactions, destinations, admin, ...)
- The `NewHandler` function has one required argument `handlersManager *routes.Manager` which is necessary to properly register endpoints and groups. 
- To create a new gin group based on proper prefix and stack of middlewares, use `handlersManager.Group` method with desired "group type" and relative path. 
- There are following "group types" available: 
  - `GroupRoot` - with no prefix and only "global" middlewares:
     - `logging.GinMiddleware(&httpLogger)`
     - `gin.Recovery()`
     - `middleware.AppContextMiddleware`
     - `middleware.CorsMiddleware`
     - `metrics.requestMetricsMiddleware`
  - `GroupOldAPI` with `/<api_version>` prefix and `auth_middleware` (besides global middlewares)
  - `GroupAPI` with `/api/<api_version> prefix and also `auth_middleware` (besides global middlewares)
  - `GroupTransactionCallback` with no prefix but with special `middleware.CallbackTokenMiddleware`
- So... for a new endpoint you'll most probably choose the `GroupAPI`
- For "registered" user enpoints you should wrap your handler with `handlers.AsUser(yourHandler)`
- Additionally, the handler func should have following arguments
```golang
func someHandler(c *gin.Context, userContext *reqctx.UserContext) 
```
- Use the `userContext` object to get XPub (XPubID) of current request.
   - Additionally, you cannot confuse "user" handlers with others - because of unique set of arguments
   
## 2. How to define an endpoint for admin
- There is no additional middleware for admin (The `auth_middleware` takes care of both `admins` and `"regular" users`
- To define a new "admin" endpoint add `handlers.AsAdmin(youAdminHandler)`, as in following example:
```golang
someGroup.POST("/access-keys/count", handlers.AsAdmin(accessKeysCount))
```
- There is no special "admin" group needed.
- But admin handlers have unique argument lists:
```golang
func someAdminHandler(c *gin.Context, _ *reqctx.AdminContext)
```
- `_ *reqctx.AdminContext` doesn't pass any information but helps to distinguish "admin" handlers from other (e.g. "user" handlers or root handlers, like "status")

- **NOTE**: Keep in mind that the signature is based on the request body, so for GET and other (no-body) requests an error will be returned
   - So add this middleware only to those endpoints which will have a BODY.
- Consider following example:
```golang
group.POST("/drafts", middleware.RequireSignature, handlers.AsUser(newTransactionDraft))
```
- There is another "not applied by default" middleware which you can add **BEFORE** actual handler to enable checking of signature.
- But if `signature_required` in **CONFIG** is `false` the signature won't be checked for `xPub` authorization method
   - For `accessKey` authorization method - the `RequireSignature` will always check the signature. 
  
## 4. How to enable signature validation for "admin" endpoint
- All "POST" and "PATCH" admin methods will check the signature by default so 
   - No explicitlly defined `RequireSignature` is needed

## 5. How to get "global" objects in a handler 
- The objects are stored in a current request context
- To get them, use those helper methods:
```golang
func someHandler(c *gin.Context) {
	engine := reqctx.Engine(c)
	logger := reqctx.Logger(c)
	config := reqctx.AppConfig(c)
}
```

# Pull Request Checklist


actions/utxos/routes.go
```diff
@@ -1,27 +1,14 @@
...
-	apiEndpoints := routes.OldAPIEndpointsFunc(func(router *gin.RouterGroup) {
-		utxoGroup := router.Group("/utxo")
-		utxoGroup.GET("", action.get)
-		utxoGroup.POST("/count", action.count)
-		utxoGroup.POST("/search", action.search)
-	})
-
-	return apiEndpoints
+func NewHandler(handlersManager *routes.Manager) {
+	roGroup := handlersManager.Group(routes.GroupOldAPI, "/utxo")
```
dorzepowski, line 10: `roGroup`? what does it mean?

Can it be `oldGroup` as in other places?

config/services.go
```diff
@@ -75,6 +75,9 @@ func (a *AppConfig) LoadTestServices(ctx context.Context) (*AppServices, error)
 	// Start services
 	_services := new(AppServices)
 
+	nopLogger := zerolog.Nop()
+	_services.Logger = &nopLogger
```
dorzepowski, line 79: Why we're disabling logger in tests? it's harder to debug tests without logger (because you always need a debugger for that)

README.md
```diff
@@ -252,7 +252,6 @@ auth:
     admin_key: xpub661MyMwAqRbcFgfmdkPgE2m5UjHXu9dj124DbaGLSjaqVESTWfCD4VuNmEbVPkbYLCkykwVZvmA8Pbf8884TQr1FgdG2nPoHR8aB36YdDQh
     require_signing: false
     scheme: xpub
-    signing_disabled: true
```
dorzepowski, line 255: I'm not sure if we should remove this option, we need to discuss this.

README.md
dorzepowski, line 1: I'm putting a comment here, although it's not related to this file.

IMHO it would be good idea, to create a `README.md` file in the `/actions` directory
and copy there a whole instruction (between `The "New Way" of handling/defining requests FAQ:` and checkboxes) into this file, to make it easier to search for description how to create endpoints.

dorzepowski (commented):
I'm still reviewing, but I already go through all the handler functions, handlers registration, README, so the most important and most hard part still left, but I want to put a comments I already made for those files.

actions/utxos/routes.go
chris-4chain: Changed to `group`.
Just for the record, It was a leftover from my "previous" idea that this is a `ReadOnlyGroup`. But I've changed the approach 🙈

config/services.go
chris-4chain: We do it almost everywhere in tests in spv-wallet. So I've just adjusted to our "standard" instead of introduced something new.
image

server/middleware/auth_middleware.go
```diff
@@ -0,0 +1,85 @@
...
+		return authByAccessKey(c, authAccessKey)
+	}
+	return nil, spverrors.ErrMissingAuthHeader
+}
+
+func authByXPub(c *gin.Context, xPub string) (*reqctx.UserContext, error) {
+	if _, err := utils.ValidateXPub(xPub); err != nil {
+		return nil, spverrors.ErrAuthorization
+	}
+	config := reqctx.AppConfig(c)
```
dorzepowski, line 45: I'm not sure about using those functions in the middleware.
It makes a hidden dependency between middlewares. And config, and logger and engine etc.

dorzepowski (changes requested):

config/services.go
dorzepowski: ok, then. Let's leave it.

server/server.go
```diff
@@ -123,93 +123,28 @@ func (s *Server) Handlers() *gin.Engine {
...
-		oldUsersAPIRoutes,
-		usersAPIRoutes,
-		// Shared Config routes
-		oldSharedConfigRoutes,
-		sharedConfigRoutes,
-	}
-
+	handlersManager := handlers.NewManager(engine, config.APIVersion)
+
+	admin.NewHandler(handlersManager)
```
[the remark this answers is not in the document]
chris-4chain: 🆗  `RegisterRoutes` is fine

server/server.go
```diff
@@ -123,93 +123,28 @@ func (s *Server) Handlers() *gin.Engine {
...
+	base.NewHandler(handlersManager)
```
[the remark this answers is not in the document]
chris-4chain: 🆗 Let's make more changes 🙈

server/handlers/wrappers.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+// AdminHandler is the handler for admin's requests
+type AdminHandler = func(c *gin.Context, _ *reqctx.AdminContext)
+
+// AsAdmin wraps the handler with the AdminContext
+func AsAdmin(handler AdminHandler) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		if c.Request.Method == "POST" || c.Request.Method == "PATCH" {
```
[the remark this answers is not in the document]
chris-4chain: That's the thing we should discuss.

server/reqctx/admincontext.go
```diff
@@ -0,0 +1,13 @@
+package reqctx
+
+// AdminContext doesn't store any significant information but help to distinguish Admin from User endpoints
+type AdminContext struct {
```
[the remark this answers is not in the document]
chris-4chain: I wanted to have `ONE` way to get the `engine` etc...
- Regardless of what kind of handler it is: admin/user/basic

server/reqctx/keys.go
```diff
@@ -0,0 +1,15 @@
...
+
+const (
+	// AppConfigKey is the key to store the app config in the request context
+	AppConfigKey = "appconfig"
+
+	// AppEngineKey is the key to store the app engine in the request context
+	AppEngineKey = "appengine"
+
+	// AppLoggerKey is the key to store the app logger in the request context
+	AppLoggerKey = "applogger"
```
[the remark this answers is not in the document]
chris-4chain: For now they do need to be public because `middleware` package uses them.
But, I could do some adjustments to make it public 🙈

server/middleware/auth_middleware.go
```diff
@@ -0,0 +1,85 @@
...
+
+	return reqctx.NewUserContextWithAccessKey(authAccessKey, accessKey, xPubObj), nil
+}
+
+func getXPubByID(c *gin.Context, xPubID string) (*engine.Xpub, error) {
+	xPubObj, err := reqctx.Engine(c).GetXpubByID(c, xPubID)
+	if err != nil || xPubObj == nil {
+		if err != nil {
+			reqctx.Logger(c).Warn().Msgf("Could not get XPub by ID: %v", err)
+		} else if xPubObj == nil {
```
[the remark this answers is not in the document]
chris-4chain: So... just `else` instead of `else if` :)
- 🆗 I'll fix it

server/middleware/auth_middleware.go
chris-4chain: And it's fine to me.

dorzepowski (changes requested):

server/handlers/wrappers.go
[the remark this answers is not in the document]
dorzepowski: client is signing all the requests, therefore maybe it's good to check how it is signed and adjust the middleware to handle it correctly.

dorzepowski (approved):

wregulski (approved):

server/server.go
```diff
@@ -116,39 +110,34 @@ func (s *Server) Handlers() *gin.Engine {
...
+	services.SpvWalletEngine.GetPaymailConfig().RegisterRoutes(ginEngine)
 
-	// Set the 404 handler (any request not detected)
-	engine.NoRoute(metrics.NoRoute, NotFound)
+	setupBaseRoutes(appConfig, ginEngine)
+}
 
-	// Set the method not allowed
-	engine.NoMethod(MethodNotAllowed)
+func setupBaseRoutes(appConfig *config.AppConfig, ginEngine *gin.Engine) {
```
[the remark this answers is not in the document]
chris-4chain: I've moved some things and got rid of this `setupBaseRoutes` 😄 
Check it out 👀

dorzepowski (changes requested):

dorzepowski (approved):

PR #687, opened 2024-09-05
feat(SPV-1020): create draft transaction handling OP RETURN output;

dorzepowski:
# Context
This is first step to prepare a new way of creating draft transactions which:
- won't be stored in a database
- will return BEEF instead of rawtx
- will return additionally some meta data of the transaction in annotations field

# How to deal with this PR

Because it's not full endpoint but only a bunch of parts that later will be used in an endpoint please look at the following parts:
1. models/request directory 
   a. contains the proposition how the json request body will look like when you want a tx with op_return
   b. it will be used also by go-client therefore I'm ensuring that the structs here can be transformed from and to JSON
2. mappings/draft - has a function that is mapping from model to more domain specific structs 
  a. it is used internally and hopefully in the future we will replace the tests for mappings here with tests of the endpoint
  b. currently I'm using just a type alias, because I don't see any benefit from rewriting it to new structure with the same fields
3. engine/transaction/....
  a. it is a main part of the changes 
  b. it is showing proposed way to handle different output types based on the op_return type.

# Pull Request Checklist


engine/transaction/draft/create_draft.go
```diff
@@ -0,0 +1,35 @@
...
+		return nil, txerrors.ErrDraftSpecificationRequired
+	}
+	outputs, annotations, err := spec.outputs(ctx)
+	if err != nil {
+		return nil, err
+	}
+	tx.Outputs = outputs
+
+	beef, err := tx.BEEFHex()
+	if err != nil {
```
wregulski, line 25: Shouldn't we validate the beef here? At least if it's not a ""? 
I would not fully rely on sdk checks as we have no control of the code there.

engine/transaction/draft/outputs/op_return.go
```diff
@@ -0,0 +1,65 @@
...
+
+// OpReturn represents an OP_RETURN output specification.
+type OpReturn opreturn.Output
+
+func (o *OpReturn) evaluate(context.Context) (annotatedOutputs, error) {
+	if len(o.Data) == 0 {
+		return nil, txerrors.ErrDraftOpReturnDataRequired
+	}
+
+	data, err := o.getData()
```
wregulski, line 24: Maybe we should iterate through this data and check if it doesn't exceed max size (80 bytes??).

engine/transaction/draft/create_op_return_draft_test.go
```diff
@@ -0,0 +1,142 @@
...
+	sdk "github.com/bitcoin-sv/go-sdk/transaction"
+	"github.com/bitcoin-sv/spv-wallet/engine/transaction"
+	"github.com/bitcoin-sv/spv-wallet/engine/transaction/draft"
+	"github.com/bitcoin-sv/spv-wallet/engine/transaction/draft/outputs"
+	"github.com/bitcoin-sv/spv-wallet/models/request/opreturn"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCreateOpReturnDraft(t *testing.T) {
+	const maxDataSize = 0xFFFFFFFF
```
wregulski, line 18: WDYT about putting this value a bit more "globally" potentially in `.../outputs/op_return.go`?

engine/transaction/draft/create_draft.go
dorzepowski: I cannot agree with that, we must start believing them and eventually either request changes/fixes or make CR with it.
We are playing to the same goal, and we want sdk to be fully reliable.

engine/transaction/draft/outputs/op_return.go
dorzepowski: you mean to validate if it doesn't exceed max size of push data to tx?
I was thinking about it, but actually the same check is done inside sdk.CreateOpReturnOutput(data) so I'm just pushing this responsibility to sdk.
If you're asking about max size of transaction, IMHO it is rather responsibility of the "broadcasting", because there is a consensus in the network of the transaction size, and how spv-wallet can know what is the current max size. Or we will be forced to maintain this size by ourselves.

engine/transaction/draft/create_op_return_draft_test.go
dorzepowski: It is a BSV related test case, and it is checked by sdk, 
I will comment this and change name, that this is a maximum data that can be pushed with OP_PUSH. But I will leave it here as otherwise it will be part of production code that is not used in production.

engine/transaction/draft/create_draft_test.go
```diff
@@ -0,0 +1,39 @@
+package draft_test
```
chris-4chain, line 1: As I recall - the package names should not contain "underscore"

models/request/opreturn/output.go
```diff
@@ -0,0 +1,10 @@
+package opreturn
+
+type Output struct {
+	DataType DataType `json:"dataType,omitempty"`
+	Data     []string `json:"data"`
+}
+
+func (o Output) GetType() string {
+	return "op_return"
```
chris-4chain, line 9: In the future, maybe, we could use some constant/enum here.

engine/transaction/draft/create_draft_test.go
dorzepowski: image
https://pkg.go.dev/testing

The `_test` package gives you a possibility to test your solution from the perspective of the client - component from different package, because it is treated as different package, but allowed to be in the same package as the other in that directory.

models/request/opreturn/output.go
dorzepowski: I'm not sure yet if this is needed/useful, those constant/enums doesn't prevent you from using custom value (for string you can assign a string, for iota you can assign any int).

engine/transaction/draft/create_draft_test.go
chris-4chain: Of course - It was indeed late hour of work that I haven't notice that 🙈

models/request/opreturn/data_type.go
```diff
@@ -0,0 +1,48 @@
...
+type DataType int
+
+const (
+	DataTypeDefault DataType = iota
+	DataTypeStrings
+	DataTypeHexes
+)
+
+// UnmarshalJSON custom unmarshaler for DataTypeEnum
+func (d *DataType) UnmarshalJSON(data []byte) error {
```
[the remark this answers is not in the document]
dorzepowski: It doesn't work for value receiver, because we cannot override the value of the value receiver.

engine/transaction/draft/create_draft.go
```diff
@@ -0,0 +1,35 @@
...
+func Create(ctx context.Context, spec *TransactionSpec) (*Transaction, error) {
+	tx := &sdk.Transaction{}
+	if spec == nil {
+		return nil, txerrors.ErrDraftSpecificationRequired
+	}
+	outputs, annotations, err := spec.outputs(ctx)
+	if err != nil {
+		return nil, err
+	}
+	tx.Outputs = outputs
```
chris-4chain, line 22: The `tx` is unnecessarily created at the top. 
It could be done here (where tx.Outputs line currently is):

```suggestion
	tx := &sdk.Transaction{
	    Outputs: outputs,
	}
```

engine/transaction/draft/outputs/op_return.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		return nil, spverrors.Wrapf(err, "failed to create OP_RETURN output")
+	}
+
+	annotation := transaction.NewDataOutputAnnotation()
+	return singleAnnotatedOutput(output, annotation), nil
+}
+
+func (o *OpReturn) getData() ([][]byte, error) {
+	data := make([][]byte, 0)
```
chris-4chain, line 41: You could preallocate the `data` slice (because the length of `o.Data` is known at this point)

engine/transaction/draft/outputs/spec.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+}
+
+func (a annotatedOutputs) splitIntoTransactionOutputsAndAnnotations() ([]*sdk.TransactionOutput, transaction.OutputsAnnotations) {
+	outputs := make([]*sdk.TransactionOutput, len(a))
+	annotations := make(transaction.OutputsAnnotations)
+	for i, out := range a {
+		outputs[i] = out.TransactionOutput
+		if out.OutputAnnotation != nil {
+			annotations[i] = out.OutputAnnotation
```
chris-4chain, line 82: It looks like possible `index out of range` but actually it's not - because it is a `map` 🙈 
Could you rename the `annotations` local variable name, that it suggests that it's not an array but rather a map?

dorzepowski: I tried to change the name, hope it will be more clear 🤷

chris-4chain (approved):

engine/transaction/draft/create_draft.go
wregulski: Okay, I won't argue :D

engine/transaction/draft/outputs/op_return.go
wregulski: Yes, I was thinking about max size of push data but if it's already in sdk it's ok

wregulski (approved):

engine/transaction/draft/create_op_return_draft_test.go
wregulski: Ok

PR #697, opened 2024-09-11
refactor(SPV-1020): move paymail servant to separate package and use it in engine

dorzepowski:
# Context
- in the next step of new create draft transaction method we need to integrate with paymail
- using paymail servant when it was in engine from the other package used by engine is ending with circular deps error
- I changed the name of PaymailServant to paymail.ServiceClient - to not have redundant Paymail in package and struct name
- most important functionality: getCapabilities with cacheing, was done by private function in engine, while it seems like a very useful method that could be used by any thing, that's why I moved it to the paymail.ServiceClient
- I moved some of the functions from paymail.go to be a methods on paymail service client and make the current code that was using them to use them with help of paymail.ServiceClient
- I moved a paymail_test.go but it has several things that was mocking paymail host server
- I created a method to setup paymail client with connection to mocked paymail host server and with methods to help to setup this paymail host server.

# Pull Request Checklist


engine/paymail/paymail_service_client.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+// GetCapabilities is a utility function to retrieve capabilities for a Paymail provider
+func (s *service) GetCapabilities(ctx context.Context, domain string) (*paymail.CapabilitiesPayload, error) {
+	// Attempt to get from cachestore
+	// todo: allow user to configure the time that they want to cache the capabilities (if they want to cache or not)
+	capabilities := new(paymail.CapabilitiesPayload)
+	if s.cs != nil {
+		if err := s.cs.GetModel(
+			ctx, cacheKeyCapabilities+domain, capabilities,
```
chris-4chain, line 85: I see you're making the cache key twice - here and at the bottom of this function.
Consider making a local variable 
```golang
cacheKey := cacheKeyCapabilities+domain
```
and reuse it everywhere you need in the function.

engine/paymail/paymail_service_client.go
```diff
@@ -0,0 +1,209 @@
...
+	case BasicPaymailPayloadFormat:
+		return "BasicPaymailPayloadFormat"
+
+	case BeefPaymailPayloadFormat:
+		return "BeefPaymailPayloadFormat"
+
+	default:
+		return fmt.Sprintf("%d", uint32(format))
+	}
+}
```
chris-4chain, line 38: This block could be moved to separate file.

engine/paymail/paymail_service_client.go
```diff
@@ -0,0 +1,209 @@
...
+
+// ServiceClient is a service that aims to make easier paymail operations.
+type ServiceClient interface {
+	GetSanitizedPaymail(addr string) (*paymail.SanitisedPaymail, error)
+	GetCapabilities(ctx context.Context, domain string) (*paymail.CapabilitiesPayload, error)
+	GetP2P(ctx context.Context, domain string) (success bool, p2pDestinationURL, p2pSubmitTxURL string, format PayloadFormat)
+	StartP2PTransaction(alias, domain, p2pDestinationURL string, satoshis uint64) (*paymail.PaymentDestinationPayload, error)
+	GetPkiForPaymail(ctx context.Context, sPaymail *paymail.SanitisedPaymail) (*paymail.PKIResponse, error)
+	AddContactRequest(ctx context.Context, receiverPaymail *paymail.SanitisedPaymail, contactData *paymail.PikeContactRequestPayload) (*paymail.PikeContactRequestResponse, error)
+}
```
chris-4chain, line 48: This interface could be moved to separate file.

engine/paymail/paymail_service_client.go
```diff
@@ -0,0 +1,209 @@
...
+
+type service struct {
+	cs cachestore.ClientInterface
+	pc paymail.ClientInterface
```
chris-4chain, line 52: Could you rename these to more descriptive names?

engine/paymail/paymail_service_client.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+// NewServiceClient creates a new paymail service client
+func NewServiceClient(cs cachestore.ClientInterface, pc paymail.ClientInterface) (ServiceClient, error) {
+	if pc == nil {
+		return nil, spverrors.Newf("paymail client is required")
```
chris-4chain, line 58: Can this happen on runtime - by design?
Or it will occur only by a developer mistake - in this case I would suggest - `panic`

engine/paymail/paymail_service_client.go
```diff
@@ -0,0 +1,209 @@
...
+			return nil, err //nolint:wrapcheck // we have handler for paymail errors
+		}
+	} else {
+		// Get the capabilities via SRV record
+		if response, err = s.pc.GetCapabilities(
+			srv.Target, int(srv.Port),
+		); err != nil {
+			return nil, err //nolint:wrapcheck // we have handler for paymail errors
+		}
+	}
```
chris-4chain, line 121: This block of code I'd suggest to move to a separate function, what returns `*paymail.CapabilitiesResponse`

engine/paymail/paymail_service_client.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// Save to cachestore
+	if s.cs != nil && !s.cs.Engine().IsEmpty() {
+		_ = s.cs.SetModel(
```
chris-4chain, line 125: I would suggest to at least log the error - but, I guess, here the logger is not available 🤔

engine/action_contact.go
```diff
@@ -16,9 +17,9 @@ func (c *Client) UpsertContact(ctx context.Context, ctcFName, ctcPaymail, reques
 		return nil, err
 	}
 
-	pmSrvnt := &PaymailServant{
-		cs: c.Cachestore(),
-		pc: c.PaymailClient(),
+	pmSrvnt, err := paymailclient.NewServiceClient(c.Cachestore(), c.PaymailClient())
+	if err != nil {
+		panic(err)
```
chris-4chain, line 22: I'm wondering if the `panic` should be inside the `NewServiceClient` function.

engine/paymail/paymail_service_client.go
dorzepowski: I'm not sure about that,
wouldn't it be easier to manage if both, interface and implementation, would be in the same file?

engine/action_contact.go
dorzepowski: hmm... maybe I will move creation of this service to be responsibility of engine and then we can panic on startup.

I need to make additional changes then, sorry 🙈

engine/paymail/paymail_service_client.go
chris-4chain: Maybe - but I'm just saying that there is such convention in this project, that the "main" interface of the service in a package is in `interface.go` file.

engine/action_contact.go
chris-4chain: Good idea 👍

wregulski:
Please assign to PR :D

wregulski (approved):

engine/action_contact.go
```diff
@@ -16,16 +17,14 @@ func (c *Client) UpsertContact(ctx context.Context, ctcFName, ctcPaymail, reques
...
 	}
 
-	pmSrvnt := &PaymailServant{
-		cs: c.Cachestore(),
-		pc: c.PaymailClient(),
-	}
-	contactPm, err := pmSrvnt.GetSanitizedPaymail(ctcPaymail)
+	paymailService := c.PaymailService()
+
+	contactPm, err := paymailService.GetSanitizedPaymail(ctcPaymail)
```
chris-4chain, line 22: If the the `Pm` stands for `PayMail` - it should be `contactPM` (according to golang styleguide)
- But I would prefer `contactPaymail`

engine/client_internal.go
```diff
@@ -127,11 +128,28 @@ func (c *Client) GetWebhooks(ctx context.Context) ([]notifications.ModelWebhook,
...
-func (c *Client) loadPaymailClient() (err error) {
+// loadPaymailComponents will load the Paymail client
+func (c *Client) loadPaymailComponents() (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			if e, ok := r.(error); ok {
+				err = spverrors.Wrapf(e, "error when creating paymail components")
+			} else {
+				err = spverrors.Newf("error when creating paymail components: %v", r)
+			}
```
chris-4chain, line 139:
```suggestion
            finalErr := r.(error) ? err : spverrors.Newf("error when creating paymail components: %v", r)
            err = spverrors.ErrorWhenCreatingPaymailComponents.Wrap()
```

Of course there is no ternary operator available in go xD

engine/paymail/paymail_payload_format.go
```diff
@@ -0,0 +1,25 @@
+package paymail
+
+import "fmt"
+
+// PayloadFormat is the format of the paymail payload
+type PayloadFormat uint32
```
chris-4chain, line 6: Are there any requirements that you've decided to use `uint32`?
Standard `int` would be "more normal xD"

engine/paymail/paymail_service_client.go
```diff
@@ -0,0 +1,194 @@
...
+
+// NewServiceClient creates a new paymail service client
+func NewServiceClient(cache cachestore.ClientInterface, paymailClient paymail.ClientInterface, log zerolog.Logger) ServiceClient {
+	if paymailClient == nil {
+		panic(spverrors.Newf("paymail client is required to create a new paymail service"))
+	}
+
+	if cache == nil {
+		log.Info().Msg("Doesn't receive cachestore, won't use cache for capabilities")
+	}
```
chris-4chain, line 31: Is it possible in our project, that we have no cache service?
Maybe during tests it is nil - but... as I recall we always have cache service (redis or freecache)

engine/paymail/paymail_service_client.go
```diff
@@ -0,0 +1,194 @@
...
+	// todo: allow user to configure the time that they want to cache the capabilities (if they want to cache or not)
+
+	cacheKey := cacheKeyCapabilities + domain
+
+	capabilities, err := s.loadCapabilitiesFromCache(ctx, cacheKey)
+	if err != nil {
+		return nil, err
+	}
+	if capabilities != nil {
+		return capabilities, err
```
chris-4chain, line 63:
```suggestion
		return capabilities, nil
```

engine/paymail/paymail_service_client.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	capabilities := new(paymail.CapabilitiesPayload)
+	err := s.cache.GetModel(ctx, key, capabilities)
+	if errors.Is(err, cachestore.ErrKeyNotFound) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, spverrors.Wrapf(err, "failed to get capabilities from cachestore")
+	}
```
chris-4chain, line 88:
```suggestion
	
	if err != nil {
	        if errors.Is(err, cachestore.ErrKeyNotFound) {
		        return nil, nil
	        }
		return nil, spverrors.Wrapf(err, "failed to get capabilities from cachestore")
	}
```

Or you did it on purpose - in that case - leave it as is 😄

engine/paymail/paymail_service_client.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func (s *service) loadCapabilities(domain string) (response *paymail.CapabilitiesResponse, err error) {
+	// Get SRV record (domain can be different!)
+	srv, err := s.paymailClient.GetSRVRecord(
+		paymail.DefaultServiceName, paymail.DefaultProtocol, domain,
+	)
+	if err != nil {
+		return
```
chris-4chain, line 114: In this case I don't see benefits of using named return values.
```suggestion
		return nil, err
```

engine/action_contact.go
dorzepowski: and what if it stand for `Paymail` not `PayMail`?

engine/client_internal.go
dorzepowski: in other words, in case if r is not error, then I will wrap "error when creating paymail components: %v" into "error when creating paymail components", doesn't seem to be a better solution

engine/paymail/paymail_payload_format.go
dorzepowski: It is done this way in engine, and I don't see good reason to change this.

dorzepowski: Man, How I miss the real enums in Go

engine/paymail/paymail_service_client.go
dorzepowski: Good question,
I don't know if this is possible, It was done this way previously so I assume that it's possible. I will try to check that.

engine/paymail/paymail_service_client.go
dorzepowski: I did it in purpose, because `errors.Is(nil, cachestore.ErrKeyNotFound)` will be false, and it's easier to reason and read without this wrapping

engine/client_internal.go
chris-4chain: Maybe I didn't mention about a key difference - 
on the caller side, my solution will allow to do this:
```
if errors.Is(err, spverrors.ErrorWhenCreatingPaymailComponents) {
 // do sth
}
```

dorzepowski: does it really matter in this case? I mean, from my point of view it should be just panic, but I'm adjusting it to current solution, where engine is eventually returning an error on creation, that's why it's this way.
For me it doesn't matter what is this error, because something go wrong in instantiation of whole engine, and actually whole application should just stop.

chris-4chain: I get it ;) 
Consider this comment resolved 👍

chris-4chain (approved):

PR #700, opened 2024-09-17
feat(SPV-1048): fixed metadata editing

Nazarii-4chain:
# Pull Request Checklist


engine/models_internal.go
```diff
@@ -97,10 +97,26 @@ func (m *Model) SetRecordTime(created bool) {
 // UpdateMetadata will update the metadata on the model
 // any key set to nil will be removed, other keys updated or added
 func (m *Model) UpdateMetadata(metadata Metadata) {
```
chris-4chain, line 99: Why don't use a "normal" assigmnent?

A deep copy or shallow copy would only be necessary if the incoming metadata were modified elsewhere after being passed. However, in our case—as far as I can tell—there is no such scenario.

Please check if my solution works:

```golang
func (m *Model) UpdateMetadata(metadata Metadata) {
	m.Metadata = metadata
}
```

chris-4chain (changes requested):

engine/models_internal.go
Nazarii-4chain: It doesn't handle case of cleared all metadata when given an empty input 
I've got this error: `panic recovered: assignment to entry in nil map` If I want to change a metadata after deleting

chris-4chain: So? Just handle the empty-map case.

Nazarii-4chain: Updated

chris-4chain: I tested my solution on my own. It worked on my side.
When have you got the `panic recovered` error. 

Did you copy paste my code:
```golang
func (m *Model) UpdateMetadata(metadata Metadata) {
	m.Metadata = metadata
}
```
which you should - or you've added some lines below?

Nazarii-4chain: With line above. Added your solution

chris-4chain (approved):

jakubmkowalski (approved):

PR #701, opened 2024-09-18
feat(SPV-1038): decouple sync job from SyncTx

chris-4chain:
image

- SyncTransaction model is no longer "used" (I left the model for now because that would delete this table from database)
- The column `tx_status` has new values (which basically replaces SyncTransaction model purpose)
- The main changes were in strategies (outgoing, internal/external incoming) and for SYNC task. 

# Pull Request Checklist


actions/transactions/broadcast_callback.go
```diff
@@ -12,21 +12,18 @@ import (
...
 		return
 	}
 
-	err = reqctx.Engine(c).UpdateTransaction(c.Request.Context(), resp)
+	err = reqctx.Engine(c).HandleTxCallback(c.Request.Context(), callbackResp)
 	if err != nil {
-		logger.Err(err).Msgf("failed to update transaction - tx: %v", resp)
-		spverrors.ErrorResponse(c, err, logger)
-		return
+		logger.Err(err).Msgf("failed to update transaction - tx: %v", callbackResp)
```
ac4ch, line 25: why here is different way of logging not like above in line 19 `spverrors.ErrorResponse(c, err, logger)`

wregulski (approved):

wregulski:
Looks good to me, all paths that I know seems covered.

engine/action_transaction.go
```diff
@@ -380,41 +354,40 @@ func (c *Client) RevertTransaction(ctx context.Context, id string) error {
...
 
-	syncTx, err := GetSyncTransactionByTxID(ctx, txInfo.ID, c.DefaultModelOptions()...)
-	if err != nil {
-		c.options.logger.Err(err).Msgf("failed to get sync transaction by tx id: %v", txInfo.ID)
+	tx.BlockHash = callbackResp.BlockHash
+	tx.BlockHeight = uint64(callbackResp.BlockHeight)
+	tx.SetBUMP(bump)
+	tx.UpdateFromBroadcastStatus(callbackResp.TxStatus)
+
+	if err := tx.Save(ctx); err != nil {
```
ac4ch, line 386: maybe is a good idea to log this error `logger.Err(err).Msgf("failed to save transaction - tx: %v", txID)`

engine/chainstate/chainstate.go
```diff
@@ -81,12 +81,7 @@ func (c *Client) QueryTransaction(
 		return nil, spverrors.ErrInvalidRequirements
 	}
```
ac4ch, line 83:
```go
if len(id) < 50 {
		return nil, spverrors.ErrInvalidTransactionID
	} else if !c.validRequirement(requiredIn) {
		return nil, spverrors.ErrInvalidRequirements
	}
```

think you should also fix this value to something like this 

```go
if len(id) < 50 {
		return nil, spverrors.ErrInvalidTransactionID
	} else if !c.validRequirement(requiredIn) {
		return nil, spverrors.ErrInvalidRequirements
	}
```

engine/paymail_service_provider.go
```diff
@@ -386,20 +386,11 @@ func saveBeefTransactionInput(ctx context.Context, c ClientInterface, input *bee
...
-
 	if bump != nil {
 		inputTx.BUMP = *bump
-		sync.SyncStatus = SyncStatusSkipped
+		inputTx.TxStatus = TxStatusMined
 	}
 
-	inputTx.syncTransaction = sync
-
 	err := inputTx.Save(ctx)
```
ac4ch, line 394: log the error

engine/tx_broadcast.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/bitcoin-sv/spv-wallet/engine/chainstate"
+	"github.com/bitcoin-sv/spv-wallet/engine/spverrors"
+)
+
+func broadcastTransaction(ctx context.Context, tx *Transaction) error {
+	client := tx.Client()
+	chainstateSrv := client.Chainstate()
+
+	defer recoverAndLog(tx.Client().Logger())
```
ac4ch, line 15: maybe this way 
```go
defer func() {
		if r := recover(); r != nil {
			tx.Client().Logger().Error().Str("txID", tx.ID).Msgf("panic in broadcastTransaction: %v", r)
			recoverAndLog(tx.Client().Logger())
		}
	}()
```

engine/tx_broadcast.go
```diff
@@ -0,0 +1,46 @@
...
+	txHex := tx.Hex
+	hexFormat := chainstate.RawTx
+	if chainstateSrv.SupportedBroadcastFormats().Contains(chainstate.Ef) {
+		// try to convert to EF, with rawRx as a fallback
+		efHex, ok := ToEfHex(ctx, tx, client)
+		if ok {
+			txHex = efHex
+			hexFormat = chainstate.Ef
+		}
+	}
```
ac4ch, line 34:
```go
if chainstateSrv.SupportedBroadcastFormats().Contains(chainstate.Ef) {
		efHex, ok := ToEfHex(ctx, tx, client)
		if ok {
			txHex = efHex
			hexFormat = chainstate.Ef
		} else {
			tx.Client().Logger().Warn().Str("txID", tx.ID).Msg("failed to convert transaction to EF format, using RawTx as fallback")
		}
	}
```

ac4ch: again log unsuccessful conversions

engine/tx_broadcast.go
```diff
@@ -0,0 +1,46 @@
...
+
+	br := chainstateSrv.Broadcast(ctx, tx.ID, txHex, hexFormat, defaultBroadcastTimeout)
+
+	if br.Failure != nil {
```
ac4ch, line 38: again i would log the failure 
```if br.Failure != nil {
		tx.Client().Logger().Error().Str("txID", tx.ID).Err(br.Failure.Error).Msg("transaction broadcast failed")
		if br.Failure.InvalidTx {
			return spverrors.ErrBroadcastRejectedTransaction.Wrap(br.Failure.Error)
		}
		return br.Failure.Error
	}
```

engine/tx_sync_task.go
```diff
@@ -0,0 +1,135 @@
...
+		return
+	}
+
+	logger.Info().Msgf("Transactions to SYNC: %d", len(txIDsToSync))
+
+	for _, record := range txIDsToSync {
+		txID := record.ID
+		tx, err := getTransactionByID(ctx, "", txID, WithClient(client))
+		if tx == nil || err != nil {
+			logger.Error().Msg("Cannot get transaction by ID even though the ID was returned from DB")
```
ac4ch, line 68: this do not explain what went wrong here 
```go
if tx == nil || err != nil {
    logger.Error().Err(err).Str("txID", txID).Msg("Cannot get transaction by ID from DB")
    continue
}
```

engine/tx_sync_task.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+
+		txInfo, err := chainstateService.QueryTransaction(ctx, txID, chainstate.RequiredOnChain, defaultQueryTxTimeout)
+
+		if err != nil {
+			switch {
+			case errors.Is(err, spverrors.ErrBroadcastUnreachable):
+				// checking subsequent transactions is pointless if the broadcast server (ARC) is unreachable will try again in the next cycle
+				logger.Warn().Msgf("%s", err.Error())
+				return
```
ac4ch, line 91: hm why we cannot `continue` here ? 

i would say:  allow subsequent transactions to be processed

engine/tx_sync_task.go
```diff
@@ -0,0 +1,135 @@
...
+		txID := record.ID
+		tx, err := getTransactionByID(ctx, "", txID, WithClient(client))
+		if tx == nil || err != nil {
+			logger.Error().Msg("Cannot get transaction by ID even though the ID was returned from DB")
+			continue
+		}
+		saveTx := func() {
+			if err := tx.Save(ctx); err != nil {
+				logger.Error().Err(err).Str("txID", txID).Msg("Cannot update transaction")
+			}
```
ac4ch, line 74: i can see lots of code like this 

```go
if err := tx.Save(ctx); err != nil {
				logger.Error().Err(err).Str("txID", txID).Msg("Cannot update transaction")
			}
```

why not create function and use it for save tx 
```go
func saveTransaction(ctx context.Context, tx *Transaction, logger *zerolog.Logger) {
    if err := tx.Save(ctx); err != nil {
        logger.Error().Err(err).Str("txID", tx.ID).Msg("Cannot update transaction")
    }
}
```

actions/transactions/broadcast_callback.go
chris-4chain: Good point 👍 But it seems to be ok, because:

Generally as @wregulski said, the ARC doesn't need to know about our internal errors.
It needs only one information - do we received the "notifiaction" or not. 

In the line `19` there is a 400 error - means that the ARC sent us wrong data in json - so it is their fault - that's why we return `status.400` and (as a side effect we log the error).

However, in the line `25` there is handling of failed processing on our side.  Means, the ARC successfully notified us about the transaction but we somehow didn't parse it well. 
- In this case we return `status.200` (you ARC sent the notification correctly) and only log a message for internal (e.g. for debugging purposes - why that happened)

engine/tx_sync_task.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"github.com/bitcoin-sv/spv-wallet/engine/chainstate"
+	"github.com/bitcoin-sv/spv-wallet/engine/spverrors"
+	"github.com/rs/zerolog"
+)
+
+// delayForBroadcastedTx indicates the time after which a broadcasted transaction should be checked
+// most probably ARC callback hasn't been received in this time so we need to check the transaction status "manually"
+func delayForBroadcastedTx() time.Time {
+	return time.Now().Add(-time.Hour)
```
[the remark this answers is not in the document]
chris-4chain: I don't thing it would make the code more readable. Moreover, the function additionally does the `time.Now().Add` stuff.
Consider this:
```golang
const delayForBroadcastedTxDuration = time.Hour
func delayForBroadcastedTx() time.Time {
	return time.Now().Add(-delayForBroadcastedTxDuration )
}
```

Is it more readable?

engine/tx_sync_task.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// delayForNotBroadcastedTx indicates the time after which a non-broadcasted transaction should be checked.
+// In this case, we don't have to wait for an ARC callback (because it will never come).
+// We're checking the transaction status after potentially enough time has passed for it to be mined.
+func delayForNotBroadcastedTx() time.Time {
+	return time.Now().Add(-10 * time.Minute)
```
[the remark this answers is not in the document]
chris-4chain: Consider my argumentation in the comment above ☝️

engine/action_transaction.go
chris-4chain: The error is actually is logged in `actions::transactions::broadcastCallback`.

Generally, I'm a fan of the approach where logging is performed at the highest possible level. For internal functions, I prefer to return errors (possibly with wrapping) and perform the logging at the topmost caller. This prevents logging an error multiple times and provides more business context about what happened.
```golang
func level3() error {
   return someErrorOnLevel3
}

func level2() error {
   if err := level3(); err != nil {
       return someErrorOnLevel2.Wrap(err)
   }
   // ...
}

// Base level
func level1() {
   if err := level2(); err != nil {
      log.Err(err)
      doSomeAction("For requests, it would return a failed status; for models, it would change some fields in the struct")
   }
}
```

engine/chainstate/chainstate.go
chris-4chain: This is not my code and it's not relevant to this task. 
Besides - this whole package will be "zaorany" soon 🙈

engine/paymail_service_provider.go
chris-4chain: The error is logged on `saveBEEFTxInputs` level.

engine/tx_broadcast.go
chris-4chain: The "not mine" function `recoverAndLog` is implemented as follows:
```golang
func recoverAndLog(log *zerolog.Logger) {
	if err := recover(); err != nil {
		log.Error().Msgf(
			"panic: %v - stack trace: %v", err,
			strings.ReplaceAll(string(debug.Stack()), "\n", ""),
		)
	}
}
```

I've just used it the same way it had been used previously.

engine/tx_broadcast.go
chris-4chain: I've just moved the function to another file - that's why it appears as "my" changes.
Additionally, I'm not sure whether or not it is necessary to log it here - it exceeds from this task context and I haven't investigate it (yet).

But, FYI - this PR is just a one step of bigger refactorization (remake) of transaction so this code most probably will be rewritten.

engine/tx_broadcast.go
chris-4chain: This "Failure" doesn't mean that something went wrong in the `spv-wallet`.
Maybe it is an invalid transaction or ARC simply doesn't know (yet) anything about this transaction. 
So, it's a part of "normal" flow and I thing the logging here is not necessary - moreover, for some cases it is actually logged, but at higher level.

engine/tx_sync_task.go
chris-4chain: This behavior has been well discussed and is valid.

The shortest explanation is in the comment above:
> Checking subsequent transactions is pointless if the broadcast server (ARC) is unreachable; will try again in the next cycle.

Consider a situation when the ARC (external) service is down for a certain amount of time.

- The task is performed periodically every `5 minutes`.
- It's pointless (and potentially dangerous) to make many calls to an unreachable server.
- Each call would last until a timeout; so, for example, with a timeout of `60 sec`, attempting to sync five transactions would cause the task execution to overlap—and that's not what we want.

engine/tx_sync_task.go
chris-4chain: I can see it too - but I'd like to minimize the refactorization of the code in this PR - because I wanted to focus on what we  need to do for now. 

However, When I'll implement the transaction repository (which it's most probably one of the next task) I would consider your suggestion.

engine/tx_sync_task.go
```diff
@@ -0,0 +1,135 @@
...
+		if tx == nil || err != nil {
+			logger.Error().Msg("Cannot get transaction by ID even though the ID was returned from DB")
+			continue
+		}
+		saveTx := func() {
+			if err := tx.Save(ctx); err != nil {
+				logger.Error().Err(err).Str("txID", txID).Msg("Cannot update transaction")
+			}
+		}
+		updateStatus := func(newStatus TxStatus) {
```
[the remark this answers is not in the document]
chris-4chain: There is one "problem" with making the separate function in the `engine` package. 
Because this `engine` is like a big sack for everything - so the function will be visible for entire package (which is huge) but applicable only in this file. 
- We're planning to split it up in the future 🙈 

Moreover I wanted to utilize the closures to minimize required arguments. 

But... when I was implementing this I - for some reason - decided to do it this way. 
But... If you are still not convinced, I'm happy to discuss the solution.

engine/tx_sync_task.go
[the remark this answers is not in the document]
wregulski: Okay, you are right, my thought was to name it to have a thought when reading:
-> okay 10mins specified here is because it's averaged time for mining a new block.

But maybe it's unnecessary

engine/tx_broadcast.go
ac4ch: ok so maybe you need extra jira ticket to change this ??? not sure

engine/chainstate/chainstate.go
ac4ch: then maybe needs separate jira to correct this ???

chris-4chain: Apparently it is `SPV-1061`

engine/tx_broadcast.go
chris-4chain: The tasks slowly appear on JIRA - but it is a brainstorming still on.

engine/tx_broadcast.go
ac4ch: so then the naming is misleading, and you should call it `invalid` not sure but anything what is marked with words invalid or failure or error in my opinion should be logged as close as possible to the check / processing tx

chris-4chain: The name is from `chainstate` which will be reworked soon (as I said above ☝️: `SPV-1061`)

However I do agree that the name is misleading - I'll try to implement/name it better in this future task. 
For now, it's out of scope of this task.

dorzepowski (changes requested):

models/errors.go
```diff
@@ -79,3 +79,11 @@ func (e SPVError) WithTrace(err error) SPVError {
 	}
 	return e.Wrap(err)
 }
+
+func (e SPVError) Is(target error) bool {
```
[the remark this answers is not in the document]
chris-4chain: yes indeed. 
Generally, I was suprised that the errors.Is(err, spverrors.SomeError) hasn't previously worked 👀 .
But I think this matter could be revised more deeply, so
- I could add a jira task to do so
- And/Or add the test cases as you proposed 

WDYT?

engine/tx_broadcast.go
chris-4chain: Maybe some new task to "Investigate potential risks of using `recoverAndLog` function ?

engine/record_tx_strategy_outgoing_tx.go
```diff
@@ -22,30 +25,49 @@ func (strategy *outgoingTx) Name() string {
...
 
 	if _shouldNotifyP2P(transaction) {
-		if err := processP2PTransaction(ctx, transaction); err != nil {
-			if revertErr := c.RevertTransaction(ctx, transaction.ID); revertErr != nil {
-				return nil, fmt.Errorf("reverting transaction failed %w; after P2P notification failed: %w", revertErr, err)
-			}
-			logger.Warn().Str("txID", transaction.ID).Msgf("processP2PTransaction failed. Reason: %v", err)
-			return nil, spverrors.ErrProcessP2PTx
+		if err = processP2PTransaction(ctx, transaction); err != nil {
+			return nil, _handleNotifyP2PError(ctx, c, transaction)
```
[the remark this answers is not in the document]
chris-4chain: I'll wrap the `cause` error 👌

engine/record_tx_strategy_outgoing_tx.go
```diff
@@ -121,23 +136,33 @@ func _hydrateOutgoingWithDraft(ctx context.Context, tx *Transaction) error {
...
+		if err := c.RevertTransaction(ctx, transaction.ID); err != nil {
+			saveAsProblematic()
+			return spverrors.ErrProcessP2PTx.Wrap(err)
 		}
+
+		// RevertTransaction saves the transaction itself as REVERTED
+		return spverrors.ErrProcessP2PTx
 	}
-	return false
+	saveAsProblematic()
```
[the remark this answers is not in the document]
chris-4chain: the function is used twice here. 
- Check the entire function `_handleNotifyP2PError`
- Consider the conditions
- If you have more readable solution - please help me

engine/tx_broadcast.go
chris-4chain: We could use "named return params" and `"catch"` the panics inside some function but return an error 

```golang
func someFunc() (err error) {
    defer func() {
        if err := recover(); err != nil {
		//optional logging
                err = spverrors.SomeError.Wrap(err)
	}
    }
    //..normal flow
   
}
```

engine/record_tx_strategy_outgoing_tx.go
```diff
@@ -22,30 +25,49 @@ func (strategy *outgoingTx) Name() string {
...
+	if err = broadcastTransaction(ctx, transaction); err != nil {
 		logger.Warn().Str("txID", transaction.ID).Msgf("broadcasting failed in outgoingTx strategy")
 		// ignore error, transaction most likely is successfully broadcasted by payment receiver
 		// TODO: return a Warning to a client
+	} else {
+		transaction.TxStatus = TxStatusBroadcasted
+	}
+
+	if err = transaction.Save(ctx); err != nil {
+		logger.Error().Str("txID", transaction.ID).Err(err).Msg("Outgoing transaction has been processed but failed save to db")
```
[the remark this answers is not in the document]
dorzepowski: ok, fine, lets keep it like now.

models/errors.go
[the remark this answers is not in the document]
dorzepowski: I would like at least to add those tests. To ensure that if someone would change anything in Is, it will still work for those cases.

engine/record_tx_strategy_outgoing_tx.go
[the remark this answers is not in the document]
dorzepowski: ok, at this stage of `engine` package I don't see any good (or better) way of doing this,
but as you see, how it looks is a little bit not readable because I didn't saw that this function was called twice

dorzepowski (approved):

engine/record_tx_strategy_outgoing_tx.go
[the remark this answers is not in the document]
dorzepowski: ⚠️  Or maybe it's better to just add this `SaveAsProblematic` to transaction model

dorzepowski (changes requested):
I forgot that I asked you to add tests for the error Is method

engine/tx_broadcast.go
chris-4chain: I've added a JIRA task for that `SPV-1068`

models/errors.go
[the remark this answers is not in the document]
chris-4chain: Done 👌

engine/record_tx_strategy_outgoing_tx.go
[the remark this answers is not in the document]
chris-4chain: `SaveAsProblematic` in transaction model would not solve the problem and purpose of this local function.
- if I did as you suggested - I would have to put the logging line twice.

dorzepowski (approved):

PR #703, opened 2024-09-19
refactor(SPV-1038): remove unused (and obsolete) query transactions fastest

dorzepowski:
# Pull Request Checklist


chris-4chain (changes requested):

engine/chainstate/chainstate.go
```diff
@@ -88,24 +88,3 @@ func (c *Client) QueryTransaction(
 	}
 	return info, nil
 }
-
-// QueryTransactionFastest will get the transaction info from ALL provider(s) returning the "fastest" valid result
-//
-// Note: this is fast but could abuse each provider based on how excessive this method is used
-func (c *Client) QueryTransactionFastest(
```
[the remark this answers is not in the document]
dorzepowski: I did it as a result of the review of your changes,
and it took me several minutes to understand that it's not used at all, 
therefore I would like to remove it to ensure that you won't reimplement it in a new shiny with ribbon still on chainstate package.

engine/chainstate/interface.go
```diff
@@ -21,9 +21,6 @@ type ChainService interface {
 	QueryTransaction(
 		ctx context.Context, id string, requiredIn RequiredIn, timeout time.Duration,
 	) (*TransactionInfo, error)
-	QueryTransactionFastest(
```
dzolt-4chain, line 24: Just out of curiosity. Why is this method unused and obsolete now? Was it inefficient or sth else?

chris-4chain: The `chainstate` was designed to use several ARC/mAPI services.
- But now we use only one ARC service - so the majority of the logic here is unnecessary.

Besides - the package will be (most probably) "zaorany" soon :D

engine/chainstate/chainstate.go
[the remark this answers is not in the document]
chris-4chain: ok 👍 some cleaning is always good

chris-4chain (approved):

engine/chainstate/interface.go
dorzepowski: it doesn't do what it's state that it's doing, because it will just use the go-broadcast-client which will try to query one by one.

PR #707, opened 2024-09-24
feat(SPV-1072): unfold and log wrapped & joined errors

chris-4chain:
### UnfoldError unfolds the error chain into a single string

It should help to debug/"get to know" what was the actual cause of some error (chain of wrapping).

example:
```
[errType1] outer error message -> [errType2] inner error message -> [errTypeN] innermost error message
```

If error message is contained by outer error, the message will be omitted, like:
```
[errType1] error message -> [errType2]
```

If error is a joined error, it will be unfolded as:
```
[type] ([errType1] joined error1 AND [errType2] joined error2)
```
For `SPVError`, it will also print the status code in parentheses:
```
[models.SPVError(404)] error message
```

### HTTP response
- Generally all errors returned as HTTP response should be typeof `SPVError`
    -  If not, a Warning with the `unfolded` error is logged (for API user - generic "Internal server error" is returned)
- For `not 5xx` errors the `Info` log level is used
    - Otherwise: `Error` log level

### zerolog error parser
- I'm proposing to set the global `zerolog.ErrorMarshalFunc` to print the `"unfolded error"`
- Thanks to this, every time you do `.Err(err)` like in: `logger.SomeLogLevel().Err(err).Msg("Some message")`
   - The detailed error info (unfolded) will be logged 

# Pull Request Checklist


dzolt-4chain:
> For not 5xx errors the Info log level is used
Otherwise: Error log level

What's the reason behind it? If we get an error that is different than 5xx wouldn't it be better to read and find in logs if it at least was a Warning level? I found it confusing a little bit at the beginning when I was looking for an error in logs but there was only INFO logs and only then I had to actually look at the response code to figure out there was indeed an error.

engine/spverrors/http_response.go
```diff
@@ -10,34 +10,40 @@ import (
...
-			Message: extendedErr.GetMessage(),
-		}, extendedErr.GetStatusCode()
+		model.Code = extendedErr.GetCode()
+		model.Message = extendedErr.GetMessage()
+		statusCode = extendedErr.GetStatusCode()
+		if statusCode >= 500 {
+			logLevel = log.Error()
+		} else {
+			logLevel = log.Info()
+		}
```
dzolt-4chain, line 38: What's the reason behind it? If we get an error that is different than 5xx wouldn't it be better to read and find in logs if it at least was a Warning level? I found it confusing a little bit at the beginning when I was looking for an error in logs but there was only INFO logs and only then I had to actually look at the response code to figure out there was indeed an error.

chris-4chain: The reason for assigning a higher log level to `5xx` status codes is that these errors typically `indicate server-side issues`, which are more likely to be caused by problems within "our system". 
- By using a higher log level (like Error) for these, we're emphasizing the need to address potential bugs or misconfigurations on our end.

On the other hand, `4xx` errors generally indicate client-side issues (e.g., incorrect passwords or malformed request data). These errors are part of normal operations, and since they're not related to bugs in our system, we log them at the Info level. 
- Logging them as warnings or errors might mislead developers or users into thinking there's an issue with the system itself, when in fact it's just standard behavior.

cmd/main.go
```diff
@@ -42,6 +43,8 @@ func main() {
 		return
 	}
 
+	spverrors.SetupGlobalZerologErrorHandler()
```
dorzepowski, line 46: shouldn't it be a part of logging package?
and why it isn't placed just after default logger (or even earlier)

engine/spverrors/unfold.go
```diff
@@ -0,0 +1,92 @@
...
+
+		if prevMsg != "" {
+			result.WriteString(" -> ")
+		}
+
+		printTypename(current, &result)
+
+		if errListUnwrapper, ok := current.(errListUnwrapper); ok {
+			unfoldJoinedErrors(errListUnwrapper, &result)
+			break // joned errors cannot be unfolded to keep this as chain of errors instead of a tree
```
dorzepowski, line 50: jo*i*ned :)

engine/spverrors/unfold.go
```diff
@@ -0,0 +1,92 @@
...
+	for current := err; current != nil; current = errors.Unwrap(current) {
+		msg := current.Error()
+
+		if prevMsg != "" {
+			result.WriteString(" -> ")
+		}
+
+		printTypename(current, &result)
+
+		if errListUnwrapper, ok := current.(errListUnwrapper); ok {
```
dorzepowski, line 48: you're overriding type name with variable name

engine/spverrors/unfold.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/bitcoin-sv/spv-wallet/models"
+)
+
+type errListUnwrapper interface {
```
dorzepowski, line 12: maybe one of:
- joinedErrors
- wrappedErrorList
- wrappedErrors

engine/spverrors/unfold.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		printTypename(current, &result)
+
+		if errListUnwrapper, ok := current.(errListUnwrapper); ok {
+			unfoldJoinedErrors(errListUnwrapper, &result)
+			break // joned errors cannot be unfolded to keep this as chain of errors instead of a tree
+		}
+
+		if prevMsg == "" || !strings.Contains(prevMsg, msg) {
```
dorzepowski, line 53: I'm not sure if this is clear, I'm rather used to have redundant message on each level then suddenly see a lack of message on the root cause of whole stack

engine/spverrors/unfold.go
```diff
@@ -0,0 +1,92 @@
...
+		return ""
+	}
+
+	prevMsg := ""
+	result := strings.Builder{}
+	for current := err; current != nil; current = errors.Unwrap(current) {
+		msg := current.Error()
+
+		if prevMsg != "" {
+			result.WriteString(" -> ")
```
dorzepowski, line 43: maybe we could consider adding new lines :see_no_evil:

dorzepowski (changes requested):

engine/spverrors/http_response.go
dorzepowski: I agree with @dzolt-4chain although it is a standard behaviour we should consider 4XX as warnings, because it should warn us that someone is misusing our system, and we (or rather maintainers of the production env) should earlier or later address those warnings, either by solving users issue or blocking them.

engine/spverrors/unfold_test.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+func TestUnfoldError(t *testing.T) {
+	t.Run("unfold single string error", func(t *testing.T) {
+		err := errors.New("test error")
+		unfolded := UnfoldError(err)
+		require.Equal(t, "[*errors.errorString] test error", unfolded)
+	})
+
+	t.Run("unfold single SPVError", func(t *testing.T) {
```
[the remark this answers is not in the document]
chris-4chain: Personally I don't like table tests :P 
Because I can "run or debug" single test case. However, convince me 😄 - why should I use table tests 🙈 ?

engine/spverrors/http_response.go
chris-4chain: So, is that ok for you?
```
if statusCode >= 500 {
	logLevel = log.Error()
} else {
	logLevel = log.Warn()
}
```

engine/spverrors/http_response.go
```diff
@@ -10,34 +10,40 @@ import (
...
 
-	logError(log, err)
-	return models.ResponseError{
-		Code:    models.UnknownErrorCode,
-		Message: "Unable to get information about error",
-	}, 500
-}
-
-func logError(log *zerolog.Logger, err error) {
 	if log != nil {
```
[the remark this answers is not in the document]
chris-4chain: Yea - I've change the code I forgot to take care of this. 
Will adjust 👍

cmd/main.go
chris-4chain: Yes and yes 👍

engine/spverrors/unfold.go
chris-4chain: It works like that, but I guess that's not good for readability 🙈 .
Will change 👍

engine/spverrors/unfold.go
chris-4chain: https://go.dev/doc/effective_go#interface-names

> By convention, one-method interfaces are named by the method name plus an -er suffix or similar modification to construct an agent noun: Reader, Writer, Formatter, CloseNotifier etc.

So, maybe - simple:
- `unwrapper`
- or as I named it
- or inline interface (like they do in `golang` std package:

https://github.com/golang/go/blob/bc047b62b2ec0c6eae3437a26410039c82365f60/src/errors/wrap.go#L62

```golang
if unwrapper, ok := current.(interface { Unwrap() []error }); ok {
```

engine/spverrors/unfold.go
chris-4chain: I thought the same but then I realized that the message can be the same on several (not both), but several levels.

Consider `pkgerrors.Wrap(err, "SOME TEXT")`
The output according to your suggestion would be:
```
[*errors.withStack] ERROR MESSAGE: SOME TEXT -> [*errors.withMessage] -> ERROR MESSAGE: SOME TEXT -> [*actualErrType] SOME TEXT
```

Take into account that there might be several levels of wrapping and the texts are usually longer than here.
My proposition would be:
```
[*errors.withStack] ERROR MESSAGE: SOME TEXT -> [*errors.withMessage] -> [*actualErrType] 
```

engine/spverrors/unfold.go
chris-4chain: zerolog would print them as `\n` and it's not - I think - you would like to get.
Besides - for production env - we'll most probably use json logger and I'm not sure if the new lines would ease the "reading" process.

cmd/main.go
chris-4chain: Done :)

engine/spverrors/http_response.go
dorzepowski: Yes exactly

engine/spverrors/unfold_test.go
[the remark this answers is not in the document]
wregulski: I agree with Krzysiek in this thread :D 
In VSCode it's painful

wregulski (approved):

engine/spverrors/unfold_test.go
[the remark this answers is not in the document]
chris-4chain: I haven't said that VSCode is painfull 🙈 I like it 😄

engine/spverrors/unfold.go
dorzepowski: ok, for now, let's keep it as is

engine/spverrors/unfold.go
dorzepowski: If we change Wrapf to use error.WithMessage, we can get rid of one level of this nesting.
Beside that, as I said, I (and probably many other developers) used to have duplicated info on many levels of the errors nesting. But at least we don't need to search for the info what was wrong, because it is on the level we search for.

engine/spverrors/unfold.go
dorzepowski: I would distinguish name of the public purpose interfaces from the helper interface that is used for some logic.
Wouldn't it show cleaner intention when you will use something like this:
``
if joined, ok := current.(joinedErrors); ok {
  unfoldJoinedErrors(joined, &result)
  break;
}
```

Comment (chris-4chain) [reply engine/spverrors/unfold.go:12]: "naming things" 😅 
Let's do it as you with 👌

File: engine/spverrors/unfold_test.go (hunk 8 above)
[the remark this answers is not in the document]
Comment (chris-4chain) [reply engine/spverrors/unfold_test.go]: I ("Chat GPT") rewrote the test file to make it as "test table" (actually a map).
So, from now I don't have ability to debug a single case 😢 
- But I don't argue at this point, because there is almost nothing to debug here. Potentially, I'll raise this issue on "more complex" tests.

File: engine/spverrors/unfold_test.go [hunk 10]
```diff
@@ -0,0 +1,54 @@
...
+func TestUnfoldError(t *testing.T) {
+	err1 := models.SPVError{Code: "test-err1", Message: "test error1", StatusCode: 404}
+	err2 := models.SPVError{Code: "test-err2", Message: "test error2", StatusCode: 400}
+	err3 := models.SPVError{Code: "test-err3", Message: "test error3", StatusCode: 500}
+
+	testCases := map[string]struct {
+		input    error
+		expected string
+	}{
+		"unfold single string error": {
```
dorzepowski, line 22: I know that it cost you a lot, but for me, it's much easier to understand the input and expected output.

dorzepowski (approved):

PR #711, opened 2024-09-24
feat(SPV-1021): add connection to bhs merkleroots endpoint

dzolt-4chain:
Discussed with @dorzepowski, in this PR we will not be introducing tests and a separate service for merkleroots. This will be a part of different task SPV-1079 and need to wait until @chris-4chain is done with changes related to chainstate.

# Pull Request Checklist


actions/merkleroots/bhs_error_response.go
```diff
@@ -0,0 +1,46 @@
...
+	"encoding/json"
+
+	"github.com/bitcoin-sv/spv-wallet/engine/spverrors"
+	"github.com/bitcoin-sv/spv-wallet/models"
+	"github.com/gin-gonic/gin"
+	"github.com/go-resty/resty/v2"
+	"github.com/rs/zerolog"
+)
+
+type bHSErrorResponse struct {
```
chris-4chain, line 13: Lowercased "b" ?

actions/merkleroots/check_bhs_status.go
```diff
@@ -0,0 +1,47 @@
...
+	if bhsConfig.URL == "" {
+		logger.Error().Msgf("%s - url not configured", errCheckingBHSMsg)
+	}
+
+	if bhsConfig.AuthToken == "" {
+		logger.Warn().Msg("warning checking Block Headers Service - auth token is not set. Some requests might not work")
+	}
+
+	res, err := httpClient.
+		SetTimeout(5 * time.Second).
```
chris-4chain, line 31: If it's a global (for this http client) timeout - do it once where you make the client.

If it's a per-request do it after `R()`:

```golang 
resp, err := client.R().
    SetTimeout(3 * time.Second).
    Get("https://example.com")
```

actions/merkleroots/get.go
```diff
@@ -0,0 +1,79 @@
...
+		return
+	}
+
+	query := url.Values{}
+	if batchSize != "" {
+		query.Add("batchSize", batchSize)
+	}
+	if lastEvaluatedKey != "" {
+		query.Add("lastEvaluatedKey", lastEvaluatedKey)
+	}
```
chris-4chain, line 38: I believe you can do this using `resty`. 

```golang
resp, err := client.R().
    SetQueryParam("key", "value").
    Get("https://example.com")
```

actions/merkleroots/get.go
```diff
@@ -0,0 +1,79 @@
...
+	bhsURL.RawQuery = query.Encode()
+
+	req := client.R().
+		EnableTrace()
+
+	if appConfig.BHS.AuthToken != "" {
+		req.SetAuthToken(appConfig.BHS.AuthToken)
+	}
+
+	res, err := req.Get(bhsURL.String())
```
chris-4chain, line 49: Did you consider if there are retries enabled for this call? 
- Are they logically necessary in this case?

actions/merkleroots/get.go
```diff
@@ -0,0 +1,79 @@
...
+		return
+	}
+
+	var response any
+	err = json.Unmarshal(res.Body(), &response)
+	if err != nil {
+		spverrors.ErrorResponse(c, spverrors.ErrBHSParsingResponse, logger)
+		return
+	}
+
```
chris-4chain, line 66: You could use resty's `SetResult` to do it for you :)

actions/merkleroots/get.go
```diff
@@ -0,0 +1,79 @@
...
+	c.JSON(http.StatusOK, response)
+}
+
+// createBHSURL parses Block Header Url from configuration and constructs a valid
+// endpoint with provided endpointPath variable
+func createBHSURL(appConfig *config.AppConfig, endpointPath string) (*url.URL, error) {
+	url, err := url.Parse(appConfig.BHS.URL + "/api/" + config.APIVersion + endpointPath)
```
chris-4chain, line 73: To me `Sprintf` will be more readable 🙈

actions/merkleroots/routes.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"github.com/bitcoin-sv/spv-wallet/config"
+	"github.com/bitcoin-sv/spv-wallet/server/handlers"
+)
+
+// RegisterRoutes creates the specific package routes
+func RegisterRoutes(appConfig *config.AppConfig, handlersManager *handlers.Manager) {
+	group := handlersManager.Group(handlers.GroupAPI, "/merkleroots")
+	group.GET("", handlers.AsUserWithAppConfig(get, appConfig))
+}
```
chris-4chain, line 12: You don't have to add another `handlers` wrapper.
The user config is already provided and can be get using:
```
reqctx.AppConfig(c)
```

For example like this:
image

server/handlers/wrappers.go
```diff
@@ -24,6 +28,19 @@ func AsUser(handler UserHandler) gin.HandlerFunc {
...
+func AsUserWithAppConfig(handler UserHandlerWithAppConfig, appConfig *config.AppConfig) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		userContext := reqctx.GetUserContext(c)
+		if userContext.GetAuthType() == reqctx.AuthTypeAdmin {
+			spverrors.AbortWithErrorResponse(c, spverrors.ErrAdminAuthOnUserEndpoint, nil)
+			return
+		}
+		handler(c, appConfig)
+	}
+}
```
chris-4chain, line 42: This is not needed see my comment above ☝️

chris-4chain (changes requested):

actions/merkleroots/bhs_error_response.go
dzolt-4chain: It's only used in this file and for parsing the response error to later parse it to spverror. I think it's unnecessary to make it exported, but I can change it if you'd like

actions/merkleroots/check_bhs_status.go
dzolt-4chain: Urm, R(). doesn't have SetTimeout method, it's used per client, I actually forgot to remove it after this request. I will add that timeout removal afterwards. Unless you've got better idea how to tackle it

actions/merkleroots/get.go
dzolt-4chain: Right. I'll change it accordingly. Will require a little different structure since those params might not be initialized

actions/merkleroots/get.go
dzolt-4chain: I didn't consider retries. Do we need them in this case? I thought the case was to actually send a request and if i.e the connection drops the client will request the server again. WDYT?

actions/merkleroots/get.go
dzolt-4chain: Right, will change it

actions/merkleroots/get.go
dzolt-4chain: This approach will actually give us an error if for some reason somebody put wrong URL in the config, i.e containing some escaped characters etc. I would leave it as it is.

actions/merkleroots/routes.go
dzolt-4chain: Oh, allright I missed that. I will change accordingly and test

actions/merkleroots/bhs_error_response.go
chris-4chain: OK. So please change it to less weird form 🙈 
```suggestion
type bhsErrorResponse struct {
```

actions/merkleroots/check_bhs_status.go
chris-4chain: I'm sorry I was suggested by ChatGPT... 
Indeed, there is no `SetTimeout` for single request.

I'm just thinking if it's a good idea to set the Timeout for the whole client.
I don't know if it's thread-safe.

actions/merkleroots/get.go
chris-4chain: I wanted to just let you know that `resty` has that ability. 
So if you find it usefull - just `SetRetryCount`

actions/merkleroots/get.go
chris-4chain: I didn't mean - "don't use url.Parse".
```suggestion
	url, err := url.Parse(fmt.Sprintf("%s/api/%s%s", appConfig.BHS.URL, config.APIVersion, endpointPath)
```
But's it's a minor - so leave it as is if you will 😄

engine/spverrors/definitions.go
```diff
@@ -375,3 +375,23 @@ var ErrBroadcastWrongBUMPResponse = models.SPVError{Message: "broadcast server r
 
 // ErrBroadcastRejectedTransaction is when broadcast server rejected transaction
 var ErrBroadcastRejectedTransaction = models.SPVError{Message: "broadcast rejected transaction", StatusCode: 400, Code: "error-broadcast-rejected-transaction"}
+
+// ////////////////////////////////// BLOCK HEADER SERVICE ERRORS
```
dorzepowski, line 379: I would suggest to move this section to separate file in package actions/merkleroots/errors

dorzepowski (changes requested):

engine/spverrors/definitions.go
```diff
@@ -375,3 +375,23 @@ var ErrBroadcastWrongBUMPResponse = models.SPVError{Message: "broadcast server r
...
+var ErrBHSUnreachable = models.SPVError{Message: "Block Header Service cannot be requested", StatusCode: 503, Code: "error-bhs-unreachable"}
+
+// ErrBHSBadURL is when creation of Block Header Service URL doesn't succeed. Probably a fault with the config file
+var ErrBHSBadURL = models.SPVError{Message: "cannot create Block Header Service url. Please check your configuration", StatusCode: 500, Code: "error-bhs-bad-url"}
+
+// ErrBHSParsingResponse is when creation of Block Header Service URL doesn't succeed. Probably a fault with the config file
+var ErrBHSParsingResponse = models.SPVError{Message: "cannot parse Block Header Service response", StatusCode: 500, Code: "error-bhs-parse-response"}
+
+// ErrBHSInvalidBatchSize is when Block Header Service request contains incorrect batch size query param
+var ErrBHSInvalidBatchSize = models.SPVError{Message: "batchSize must be 0 or a positive integer", StatusCode: 400, Code: "error-bhs-invalid-batch-size"}
```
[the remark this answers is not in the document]
dzolt-4chain: allright, removed BHS

engine/spverrors/definitions.go
```diff
@@ -375,3 +375,23 @@ var ErrBroadcastWrongBUMPResponse = models.SPVError{Message: "broadcast server r
...
+
+// ErrBHSMerkleRootNotFound is when Block Header Service cannot find requested merkleroot in lastEvaluatedKey query param
+var ErrBHSMerkleRootNotFound = models.SPVError{Message: "No block with provided merkleroot was found", StatusCode: 404, Code: "error-bhs-merkleroot-not-found"}
+
+// ErrBHSMerkleRootNotInLC is when Block Header Service finds merkleroot in lastEvaluateKey query param but it is not in longest chain
+var ErrBHSMerkleRootNotInLC = models.SPVError{Message: "Provided merkleroot is not part of the longest chain", StatusCode: 409, Code: "error-bhs-merkleroot-not-part-of-lc"}
```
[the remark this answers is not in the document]
dzolt-4chain: changed to longest-chain

server/handlers/wrappers.go
dzolt-4chain: removed

engine/spverrors/definitions.go
```diff
@@ -375,3 +375,23 @@ var ErrBroadcastWrongBUMPResponse = models.SPVError{Message: "broadcast server r
...
+var ErrBHSBadURL = models.SPVError{Message: "cannot create Block Header Service url. Please check your configuration", StatusCode: 500, Code: "error-bhs-bad-url"}
+
+// ErrBHSParsingResponse is when creation of Block Header Service URL doesn't succeed. Probably a fault with the config file
+var ErrBHSParsingResponse = models.SPVError{Message: "cannot parse Block Header Service response", StatusCode: 500, Code: "error-bhs-parse-response"}
+
+// ErrBHSInvalidBatchSize is when Block Header Service request contains incorrect batch size query param
+var ErrBHSInvalidBatchSize = models.SPVError{Message: "batchSize must be 0 or a positive integer", StatusCode: 400, Code: "error-bhs-invalid-batch-size"}
+
+// ErrBHSMerkleRootNotFound is when Block Header Service cannot find requested merkleroot in lastEvaluatedKey query param
+var ErrBHSMerkleRootNotFound = models.SPVError{Message: "No block with provided merkleroot was found", StatusCode: 404, Code: "error-bhs-merkleroot-not-found"}
```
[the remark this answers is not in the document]
dzolt-4chain: removed

actions/merkleroots/bhs_error_response.go
```diff
@@ -0,0 +1,43 @@
...
+	default:
+		spvErr := models.SPVError{
+			Message:    responseErr.Message,
+			StatusCode: res.StatusCode(),
+			Code:       responseErr.Code,
+		}
+		err = spverrors.ErrInternal.Wrap(spvErr)
+	}
+
+	return err
```
chris-4chain, line 42:
```suggestion
	if err := json.Unmarshal(res.Body(), &responseErr); err {
		return ErrBHSParsingResponse.Wrap(err)
	}

	switch responseErr.Code {
	case "ErrInvalidBatchSize":
		return ErrInvalidBatchSize
	case "ErrMerkleRootNotFound":
		return ErrBHSMerkleRootNotFound
	case "ErrMerkleRootNotInLC":
		return ErrBHSMerkleRootNotInLongestChain
	default:
		spvErr := models.SPVError{
			Message:    responseErr.Message,
			StatusCode: res.StatusCode(),
			Code:       responseErr.Code,
		}
		return spverrors.ErrInternal.Wrap(spvErr)
	}
```

actions/merkleroots/errors.go
```diff
@@ -0,0 +1,23 @@
+package merkleroots
+
+import "github.com/bitcoin-sv/spv-wallet/models"
+
+// ////////////////////////////////// BLOCK HEADER SERVICE ERRORS
```
dorzepowski, line 5: It's not needed here

actions/merkleroots/get_merkleroots_from_bhs.go
```diff
@@ -0,0 +1,61 @@
...
+// endpoint with provided endpointPath variable
+func createBHSURL(appConfig *config.AppConfig, endpointPath string, logger *zerolog.Logger) (*url.URL, error) {
+	if appConfig.BHS.URL == "" {
+		logger.Error().Msgf("create Block Header Service URL - url not configured")
+	}
+	if appConfig.BHS.AuthToken == "" {
+		logger.Warn().Msg("warning creating Block Headers Service url - auth token is not set. Some requests might not work")
+	}
+
+	url, err := url.Parse(appConfig.BHS.URL + "/api/" + appConfig.BHS.APIVersion + endpointPath)
```
dorzepowski, line 55: Please, no.
Use just a static string `/api/v1` 
there is no reasonable reason for providing APIVersion as a config param.
1. we're integrated with api V1, 
2. even if there will be `/api/v2` are you sure that we're ready to use it? Acutally it shouldn't because v2 means that it's not backward compatible
3. this additional property, makes the config more bloated without any benefit of giving now a possiblity for choosing api version, as there is only one BHS api version.

config.example.yaml
```diff
@@ -91,6 +91,8 @@ block_headers_service:
   auth_token: mQZQ6WmxURxWz5ch
   # URL used to communicate with Block Headers Service (BHS)
   url: http://localhost:8080
+  # api version of Block Headers Service api to use
+  api_version: v1
```
dorzepowski, line 95: as in the comment in code, please remove it.

actions/merkleroots/get_merkleroots_from_bhs.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// createBHSURL parses Block Header Url from configuration and constructs a valid
+// endpoint with provided endpointPath variable
+func createBHSURL(appConfig *config.AppConfig, endpointPath string, logger *zerolog.Logger) (*url.URL, error) {
+	if appConfig.BHS.URL == "" {
+		logger.Error().Msgf("create Block Header Service URL - url not configured")
+	}
+	if appConfig.BHS.AuthToken == "" {
+		logger.Warn().Msg("warning creating Block Headers Service url - auth token is not set. Some requests might not work")
```
dorzepowski, line 52: ### minor
This log should be rather in the else branch in the place where you're setting a header with this token.

dorzepowski (changes requested):

actions/merkleroots/get_merkleroots_from_bhs.go
dzolt-4chain: moved

actions/merkleroots/get_merkleroots_from_bhs.go
dzolt-4chain: okay, noted and changed to static string

config.example.yaml
dzolt-4chain: removed

actions/merkleroots/get.go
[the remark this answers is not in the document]
dzolt-4chain: replaced by a static string instead

chris-4chain (approved):

dorzepowski (approved):

PR #712, opened 2024-09-25
feat(SPV-1071): handle sender paymail in paymail outputs

dorzepowski:
## Some context

1. I prepared a little bit weird paymailaddress Service which is accepting some functions, it is a first step and kinda prosthesis for moving paymail addresses into own package. I'm opened for suggestion how to do it in a better way.
2. I'm introducing a test abilities pattern, which is extending test fixtures pattern with option to also provide assertions (not included yet - will be done in a separate PR). This pattern should abstract most setup of the test into an understandable methods / functions that after call should prepare known default environment for tests and provide a readable way to change this default a little bit to show a different test case scenario.


engine/client_internal.go
```diff
@@ -155,10 +156,46 @@ func (c *Client) loadPaymailComponents() (err error) {
 	return
 }
 
+func (c *Client) loadPaymailAddressService() error {
+	if c.options.paymailAddressService == nil {
```
chris-4chain, line 160: Maybe explicit return nil error if the service is already created? ...and skip the wrapping? 🤔 
```
if c.options.paymailAddressService != nil {
    return nil
```
//...rest of your code
return nil

wregulski (approved):

engine/transaction/draft/create_draft_test.go
```diff
@@ -4,41 +4,53 @@ import (
...
-			spec: &draft.TransactionSpec{Outputs: outputs.NewSpecifications()},
+			spec: &draft.TransactionSpec{
+				XPubID:  fixtures.Sender.XPubID,
+				Outputs: outputs.NewSpecifications(),
+			},
+			expectedError: txerrors.ErrDraftRequiresAtLeastOneOutput,
 		},
 	}
 	for name, test := range errorTests {
 		t.Run("return error "+name, func(t *testing.T) {
```
chris-4chain, line 42: As we discussed - Maybe without the strings concatenating here 🤔

engine/transaction/draft/draft_service.go
```diff
@@ -13,19 +14,25 @@ import (
...
 // NewDraftService creates a new draft service.
-func NewDraftService(paymailService paymail.ServiceClient, logger zerolog.Logger) Service {
+func NewDraftService(paymailService paymail.ServiceClient, paymailAddressService paymailaddress.Service, logger zerolog.Logger) Service {
 	if paymailService == nil {
-		panic("paymail service is required to create draft transaction service")
+		panic("paymail.ServiceClient is required to create draft transaction service")
+	}
+
+	if paymailAddressService == nil {
+		panic("paymailaddress.Service is required to create draft transaction service")
```
chris-4chain, line 29: I'm not sure if the panics should be here.
This function is used in `loadTransactionDraftService` which could return some error on fail.
For now it returns always `nil`.

engine/transaction/draft/outputs/paymail.go
```diff
@@ -59,7 +66,49 @@ func (p *Paymail) createBsvPaymailOutput(output *paymail.PaymentOutput, referenc
...
+func (p *Paymail) sender(ctx evaluation.Context) (string, error) {
+	if p.From != nil {
+		err := p.validateProvidedSenderPaymail(ctx)
+		if err != nil {
+			return "", err
+		}
+
+		return *p.From, nil
+	}
+	return p.defaultSenderAddress(ctx)
```
chris-4chain, line 84:
```suggestion
	if p.From == nil {
	    return p.defaultSenderAddress(ctx)
	}
        err := p.validateProvidedSenderPaymail(ctx)
        if err != nil {
	        return "", err
        }
        return *p.From, nil	
```

engine/transaction/draft/outputs/paymail.go
```diff
@@ -59,7 +66,49 @@ func (p *Paymail) createBsvPaymailOutput(output *paymail.PaymentOutput, referenc
...
+	ownsPaymail, err := ctx.PaymailAddressService().HasPaymailAddress(ctx, ctx.XPubID(), sender)
+	if errors.Is(err, spverrors.ErrCouldNotFindPaymail) {
+		return txerrors.ErrSenderPaymailAddressIsInvalid.Wrap(err)
+	}
+	if err != nil {
+		return spverrors.Wrapf(err, "failed to check if paymail %s belongs to xpub %s", sender, ctx.XPubID())
+	}
+
+	if !ownsPaymail {
+		return txerrors.ErrSenderPaymailAddressIsInvalid
```
chris-4chain, line 102: I don't know if it's necessary, but, maybe, we could distinguish the situation when `ErrCouldNotFindPaymail` and `!ownsPaymail`.

But, perhaps it's ok because of the wrapping, so for the not-find situation we can always do `errors.Is(err, ErrCouldNotFindPaymail)`

engine/transaction/errors/errors.go
```diff
@@ -6,6 +6,9 @@ var (
 	// ErrDraftSpecificationRequired is returned when a draft is created with no specification.
 	ErrDraftSpecificationRequired = models.SPVError{Code: "draft-spec-required", Message: "draft requires a specification", StatusCode: 400}
 
+	// ErrDraftSpecificationXPubIDRequired is returned when a draft is created without xPubId.
```
chris-4chain, line 9:
```suggestion
	// ErrDraftSpecificationXPubIDRequired is returned when a draft is created without xPubID.
```

engine/tester/logger.go
```diff
@@ -1,8 +1,18 @@
...
 
 // Logger returns a logger that can be used as a dependency in tests.
-func Logger() zerolog.Logger {
+func Logger(t testing.TB) zerolog.Logger {
+	return zerolog.New(zerolog.NewConsoleWriter(zerolog.ConsoleTestWriter(t)))
+}
+
+// NopLogger returns a logger that does nothing.
+// Deprecated: Use Logger instead.
+func NopLogger() zerolog.Logger {
```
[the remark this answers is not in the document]
dorzepowski: You're right

engine/transaction/draft/draft_service.go
dorzepowski: Actually after rethinking previous approaches, I think that actually we should always panic in such case in NewXyz()
It shouldn't happen at all, and if set in wrong way, than it is 100% fault of programmer and he need to fix it as soon as possible.

I would like even change all the loadXyz functions in client_internal to panic but I think that it is again just a tip of iceberg.

engine/transaction/draft/outputs/paymail.go
dorzepowski: good point

engine/transaction/draft/draft_service.go
chris-4chain: So ... maybe we should remove an option to return an error from `loadTransactionDraftService` - and perhaps for siilars 🤔 ?

engine/transaction/draft/outputs/paymail.go
dorzepowski: From the client point of view, I'm not sure if this is really helpful information that he is trying to use someone else paymail.

chris-4chain: Ok. Fine for me 👍

engine/transaction/draft/draft_service.go
dorzepowski: maybe in the future 🙈 
Or I will try to do that just after PR with custom assertions.

chris-4chain (approved):

PR #715, opened 2024-09-27
feat(SPV-1074): new chain query 

chris-4chain:
This pull request introduces a significant refactor to integrate ARC services into the SPV wallet application. The most important changes include the addition of a new chain service, modifications to transaction callback handling, and updates to configuration and testing to support the new service.

### Integration of ARC Services:

* **New Chain Service:**
  - Added `chainService` and related interfaces to handle ARC queries (`engine/chain/chain_service.go`, `engine/chain/interface.go`). [[1]](diffhunk://#diff-31c2f462dcc0321b0c576079b4309bf2e50f720333016ad708804a719e2bf03aR1-R18) [[2]](diffhunk://#diff-09b52f13f43dbaefbf98d9c12583aab9e3750313cc87b043f4e5e3401e351c15R1-R17)
  - Implemented `queryService` for querying transaction statuses from ARC (`engine/chain/internal/query/query_service.go`).
  - Added models for ARC transaction info and errors (`engine/chain/models/tx_info.go`, `engine/chain/models/arc_error.go`, `engine/chain/models/txstatus.go`). [[1]](diffhunk://#diff-83de54cafa4a21e72ea640f7411ad32a8685bc9a79e6a514bea06ded0ca3c994R1-R17) [[2]](diffhunk://#diff-9371b585912f7df8eef31fa3e3bfe84a1c0f19bbe88af0450f8d7a21e7b52e9bR1-R24) [[3]](diffhunk://#diff-d539578a8fafec3b39a80c1b1b93c4b57652d85bd4efeb278bfab3684b0ef741R1-R34)

### Modifications to Transaction Handling:

* **Transaction Callback Handling:**
  - Updated transaction callback structures to use `chainmodels.TXInfo` instead of `broadcast.SubmittedTx` (`actions/transactions/broadcast_callback.go`, `engine/action_transaction.go`). [[1]](diffhunk://#diff-b91bfba755c04aaee870baa9b6c8573b48511e760216efa819e38e208b3d3ccdL15-R15) [[2]](diffhunk://#diff-6683c0f7390c750a683b89c026a6f66650a47fb79e274023401b70dcd13d6c57L365-R365)
  - Changed import paths to reflect new ARC models (`actions/transactions/broadcast_callback.go`, `engine/action_transaction.go`, `engine/client.go`). [[1]](diffhunk://#diff-b91bfba755c04aaee870baa9b6c8573b48511e760216efa819e38e208b3d3ccdL6-R6) [[2]](diffhunk://#diff-6683c0f7390c750a683b89c026a6f66650a47fb79e274023401b70dcd13d6c57L8-R8) [[3]](diffhunk://#diff-a92e280e12097c7a058672dc80551529f7dfed8e9001beeff1109d518a0e89d0R9)

### Configuration and Testing Updates:

* **Configuration Enhancements:**
  - Added ARC configuration options and initialized chain service in the client (`engine/client.go`, `engine/client_internal.go`). [[1]](diffhunk://#diff-a92e280e12097c7a058672dc80551529f7dfed8e9001beeff1109d518a0e89d0R53-R66) [[2]](diffhunk://#diff-a92e280e12097c7a058672dc80551529f7dfed8e9001beeff1109d518a0e89d0R214-R215) [[3]](diffhunk://#diff-921c3dfbdb1e22318f59271adcea08e24b27734cd3f05f2fc8aa707f4c66e236R205-R211)
  - Updated service loading to include ARC settings (`config/services.go`).

* **Testing:**
  - Added unit tests for the query service and mock responses for ARC API (`engine/chain/internal/query/query_service_test.go`, `engine/chain/internal/query/arc_mock_test.go`). [[1]](diffhunk://#diff-269d54f5b170c180b918a38c2f1641f57fc81bbf91328cd8b973c1232e2af602R1-R131) [[2]](diffhunk://#diff-3ade3296582e3ecb45bd73a1d0468855e60ea3694d4c7195ba9fecc2178764f1R1-R65)

### Linter Configuration:

* **Linter Updates:**
  - Added `bodyclose` linter to `.golangci-lint.yml` to ensure proper handling of HTTP response bodies.

# Pull Request Checklist


engine/chain/internal/query/query_service.go
```diff
@@ -0,0 +1,106 @@
...
+		logger:       logger,
+		httpClient:   httpClient,
+		url:          url,
+		token:        token,
+		deploymentID: deploymentID,
+	}
+}
+
+// Query a transaction.
+func (s *Service) Query(ctx context.Context, txID string) (*chainmodels.TXInfo, error) {
```
dzolt-4chain, line 38: Maybe something like QueryTransaction()?

engine/chain/internal/query/query_service.go
```diff
@@ -0,0 +1,106 @@
...
+	if !s.validateTX(txID) {
+		return nil, spverrors.ErrInvalidTransactionID
+	}
+	req := s.httpClient.R().
+		SetContext(ctx).
+		SetHeader("Content-Type", "application/json").
+		SetResult(&chainmodels.TXInfo{}).
```
dzolt-4chain, line 45: What do you think about creating the TXInfo{} variable above and in line 68 refer to that variable instead of casting the result?

engine/chain/internal/query/query_service.go
```diff
@@ -0,0 +1,106 @@
...
+
+	switch response.StatusCode() {
+	case http.StatusOK:
+		return response.Result().(*chainmodels.TXInfo), nil
+	case http.StatusUnauthorized, http.StatusForbidden:
+		return nil, s.withArcError(response, spverrors.ErrARCUnauthorized)
+	case http.StatusNotFound:
+		if _, ok := s.asARCError(response); ok {
+			// ARC returns 404 when transaction is not found
+			return nil, nil
```
dzolt-4chain, line 74: why return nil, as not found error? don't we want to return an error here aswell?

engine/chain/internal/query/query_service.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, spverrors.ErrARCUnreachable
+	case http.StatusConflict:
+		return nil, s.withArcError(response, spverrors.ErrARCGenericError)
+	}
+
+	return nil, s.withArcError(response, spverrors.ErrARCUnsupportedStatusCode)
+}
+
+func (s *Service) validateTX(txID string) bool {
+	return len(txID) >= 50
```
dzolt-4chain, line 85: maybe create a constant from 50 or add a comment? I don't understand why 50 and some might neither

engine/chain/internal/query/query_service_test.go
```diff
@@ -0,0 +1,131 @@
...
+				require.Equal(t, tc.expectStatus, string(txInfo.TXStatus))
+			}
+		})
+	}
+}
+
+type timeoutDst int
+
+const (
+	applyTimeoutCtx timeoutDst = iota + 1
```
dzolt-4chain, line 129: what's the reason behind typing it above and not using int in a const? Also what does the iota assignment mean that applyTimeoutCtx is equal to 1 and applyTimeoutResty is equal to 2?

engine/chain/models/arc_error.go
```diff
@@ -0,0 +1,24 @@
+package chainmodels
+
+import "fmt"
+
+// ArcError represents an error returned by the ARC API when status code is 4xx.
```
dzolt-4chain, line 5: Do we also want to consider handling other errors than 4xx? If not could you tell me why we only handle 4xx structure here?

engine/spverrors/definitions.go
```diff
@@ -375,3 +375,18 @@ var ErrBroadcastWrongBUMPResponse = models.SPVError{Message: "broadcast server r
 
 // ErrBroadcastRejectedTransaction is when broadcast server rejected transaction
 var ErrBroadcastRejectedTransaction = models.SPVError{Message: "broadcast rejected transaction", StatusCode: 400, Code: "error-broadcast-rejected-transaction"}
+
+// ErrARCUnreachable is when ARC cannot be requested
+var ErrARCUnreachable = models.SPVError{Message: "ARC cannot be requested", StatusCode: 404, Code: "error-broadcast-unreachable"}
```
dzolt-4chain, line 380: maybe code 500 here? I'm not sure about this but 404 might indicate that we tried an endpoint that doesn't exist. WDYT?

engine/spverrors/definitions.go
```diff
@@ -375,3 +375,18 @@ var ErrBroadcastWrongBUMPResponse = models.SPVError{Message: "broadcast server r
...
+
+// ErrARCUnauthorized is when ARC returns unauthorized
+var ErrARCUnauthorized = models.SPVError{Message: "ARC returned unauthorized", StatusCode: 401, Code: "error-broadcast-unauthorized"}
+
+// ErrARCParseResponse is when ARC response cannot be parsed
+var ErrARCParseResponse = models.SPVError{Message: "ARC response cannot be parsed", StatusCode: 400, Code: "error-broadcast-parse-response"}
```
dzolt-4chain, line 386: Here perhaps 500 aswell, as 400 indicates there was something wrong with the request and not the response. WDYT?

dorzepowski (changes requested):

engine/chain/internal/query/query_service.go
chris-4chain: Lets stick to the previous name - I'm gonna rename to `QueryTransaction`

engine/chain/internal/query/query_service.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+		return nil, spverrors.ErrARCUnreachable
+	case http.StatusConflict:
+		return nil, s.withArcError(response, spverrors.ErrARCGenericError)
+	}
+
+	return nil, s.withArcError(response, spverrors.ErrARCUnsupportedStatusCode)
+}
+
+func (s *Service) validateTX(txID string) bool {
```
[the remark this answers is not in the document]
chris-4chain: That was copied from our old chainstate code...
I agree, we could validate the txID - but should we @dorzepowski ?

engine/chain/internal/query/arc_mock_test.go
```diff
@@ -0,0 +1,65 @@
...
+const arcURL = "https://api.testarc.com/arc"
+const arcToken = "mock-token"
+
+func arcMockActivate() (*resty.Client, deactivationMock) {
+	client := resty.New()
+	httpmock.ActivateNonDefault(client.GetClient())
+
+	responder := func(status int, content string) func(req *http.Request) (*http.Response, error) {
+		return func(req *http.Request) (*http.Response, error) {
+			time.Sleep(100 * time.Millisecond)
```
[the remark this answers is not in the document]
chris-4chain: Show me the way as simple as my way - so I would apply it.

engine/chain/internal/query/query_service_test.go
```diff
@@ -0,0 +1,131 @@
...
+				var cancel context.CancelFunc
+				ctx, cancel = context.WithTimeout(ctx, 1*time.Millisecond)
+				defer cancel()
+			} else if tc.applyTimeout == applyTimeoutResty {
+				service.httpClient.SetTimeout(1 * time.Millisecond)
+			}
+
+			txInfo, err := service.Query(ctx, tc.txID)
+
+			if tc.expectErr != nil {
```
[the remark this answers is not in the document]
chris-4chain: To be honest I've initially implemented it without test-table. But I've forced myself to align with your suggestion to use test-table. 

Splitting into `success cases` and `error cases` won't solve the problem, because for `error cases` I would have to define `timeout errors cases` and `other error cases`.

Taking above into account I would go back to standard test cases without test-table. Agree @dorzepowski ?

engine/chain/internal/query/query_service_test.go
```diff
@@ -0,0 +1,131 @@
...
+		t.Run(name, func(t *testing.T) {
+			httpClient, reset := arcMockActivate()
+			defer reset()
+
+			service := NewQueryService(logger, httpClient, tc.arcURL, tc.arcToken, deploymentID)
+
+			ctx := context.Background()
+			if tc.applyTimeout == applyTimeoutCtx {
+				var cancel context.CancelFunc
+				ctx, cancel = context.WithTimeout(ctx, 1*time.Millisecond)
```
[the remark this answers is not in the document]
chris-4chain: The zeros are always tricky - I'm not sure how that would behave - I'd like not to do that.

engine/chain/internal/query/query_service_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			httpClient, reset := arcMockActivate()
+			defer reset()
+
+			service := NewQueryService(logger, httpClient, tc.arcURL, tc.arcToken, deploymentID)
+
+			ctx := context.Background()
+			if tc.applyTimeout == applyTimeoutCtx {
```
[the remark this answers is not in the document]
chris-4chain: Consider my suggestion to go back to non-table-tests ☝️

engine/chain/chain_service.go
```diff
@@ -0,0 +1,18 @@
+package chain
+
+import (
+	"github.com/bitcoin-sv/spv-wallet/engine/chain/internal/query"
+	"github.com/go-resty/resty/v2"
+	"github.com/rs/zerolog"
+)
+
+type chainService struct {
```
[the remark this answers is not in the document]
chris-4chain: The reason is one:
- not only QueryService will be in the ChainService
- For now it is, but in near future there will be:
```golang
type chainService struct {
	QueryService
	BroadcastService
	YourMerkleRootsService
}
```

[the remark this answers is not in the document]
dzolt-4chain: ahh allright if that's the case

engine/chain/internal/query/query_service.go
chris-4chain: In documentation they mention both ways. 
https://pkg.go.dev/github.com/go-resty/resty/v2#Request.SetResult

I think we should be consistent how we do this so I think we should decide and stick to this. 

So... I don't know 
- Reffering to a pointer to initially defined empty struct is more straight forward
- The `Result` seems to be more readable
- But the casting is ugly
- But in more sophisticated scenario when we set different models under some conditions, The `Result()` would be more accurate.
```golang
req := s.httpClient.R().
if someCondition {
    req.SetResponse(&SomeModel{})
}else {
    req.SetResponse(&OtherModel{})
}

response, err := req.Get("url")
switch res := response.Result().(type) {
    case *SomeModel:
         return handleSomeModelRes(res)
   case *OtherModel:
         return  handleOtherModelRes(res)
}

```

engine/chain/internal/query/query_service_test.go
chris-4chain: that's "normal" enum in golang 🙈

engine/chain/models/arc_error.go
chris-4chain: That's the only one model from ARC documentation. 
They always returns it when "something-went-wrong".

So what other models you would suggest to handle?

engine/spverrors/definitions.go
chris-4chain: The status code is not-applicable for this error.
This is why we've recently been discussing how to refactor it.
But sure 👍 
500 would look better - but, whatever 😄

engine/chain/chain_service.go
```diff
@@ -0,0 +1,18 @@
...
+	QueryService
+}
+
+// NewChainService creates a new chain service.
+func NewChainService(logger zerolog.Logger, arcURL, arcToken, deploymentID string) Service {
+	return &chainService{
+		query.NewQueryService(logger.With().Str("chain", "query").Logger(), resty.New(), arcURL, arcToken, deploymentID),
```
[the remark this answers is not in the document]
chris-4chain: Maybe I would do that but please let's postpone it to the following PRs when we'll have more that one service (e.g. two xD)

engine/chain/internal/query/query_service.go
chris-4chain: This is similar to when you query Database but there was no result.
- Database is reachable
- Database responded with valid response
- Bot now rows are in the response

engine/chain/models/arc_error.go
dzolt-4chain: Ah okay, I thought they also return some "different" errors for example when a 5xx is thrown or something. If it's the only model than it's all good

engine/chain/internal/query/query_service.go
chris-4chain: OK :D 
Let's stick to the simpler method with "pointer referral".

Maybe in the future we'll revisit this consideration 😄

dzolt-4chain (approved):

engine/chain/chain_service.go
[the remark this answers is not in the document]
dorzepowski: But this will enable us to create a service for merkleroots,
one of the reason we postponed that, it was to prevent reimplementing shared resty for each.

engine/chain/internal/query/query_service.go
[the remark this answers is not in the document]
dorzepowski: I don't know where we use it. 
If we don't want to validate, then just don't, 
if we want to validate, then we should rather use go-sdk

engine/chain/internal/query/query_service_test.go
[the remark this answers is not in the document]
dorzepowski: it depends,
if you have single positive case, then feel free to keep it as standard test
if you have single error case then feel free to keep it as single standard test
if you have single timeout case then feel free to keep it as single standard test

But if you have several test cases, that actually are exactly the same, just input and output is changing, then it's better to have table tests.
IMHO we should pick a tool that the best fits our needs, I don't see any reason to create table tests for single test case.

engine/chain/chain_service.go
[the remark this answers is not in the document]
chris-4chain: ok, ok 👍 Let's do this :)

engine/chain/internal/query/query_service.go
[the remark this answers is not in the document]
chris-4chain: I've removed the validation at this point.
Let's allow ARC validate 👍 
- I've checked the response (with actual ARC server) and adjusted a test case.

engine/chain/internal/query/arc_mock_test.go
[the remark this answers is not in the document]
chris-4chain: I've introduced `applyTimeout` func argument to conditionally apply this sleep.

engine/chain/internal/query/query_service_test.go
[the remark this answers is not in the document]
chris-4chain: I've reorganized things - check it out 👍

dorzepowski (approved):

PR #722, opened 2024-10-04
refactor(SPV-1071): cleanup in config package.

dorzepowski:
# Pull Request Checklist


config.example.yaml
```diff
@@ -121,3 +119,9 @@ task_manager:
 # Prometheus metrics configuration
 metrics:
   enabled: false
+
```
wregulski, line 122: IMO there is unnecessary line 122

logging/logging.go
```diff
@@ -15,8 +16,21 @@ const (
 	jsonLogFormat    = "json"
 )
 
-// CreateLogger create and configure zerolog logger based on app config.
-func CreateLogger(instanceName, format, level string, logOrigin bool) (*zerolog.Logger, error) {
+// GetDefaultLogger create and configure default zerolog logger. It should be used before config is loaded
+func GetDefaultLogger() zerolog.Logger {
+	logger, err := createLogger("spv-wallet-default", jsonLogFormat, "debug", true)
```
wregulski, line 21: Should we have `jsonLogFormat` here? I would vote for `consoleLogFormat` as a default 🤔

cmd/main.go
```diff
@@ -10,19 +10,22 @@ import (
...
 
 	"github.com/bitcoin-sv/spv-wallet/config"
-	"github.com/bitcoin-sv/spv-wallet/dictionary"
 	_ "github.com/bitcoin-sv/spv-wallet/docs"
+	"github.com/bitcoin-sv/spv-wallet/engine"
 	"github.com/bitcoin-sv/spv-wallet/logging"
 	"github.com/bitcoin-sv/spv-wallet/server"
 )
 
+// version version of the application that can be overridden with ldflags during build
```
chris-4chain, line 19: version word is twice 😊

config/config_to_options.go
```diff
@@ -0,0 +1,298 @@
...
+			callbackToken, err := utils.HashAdler32(DefaultAdminXpub)
+			if err != nil {
+				return nil, spverrors.Wrapf(err, "error while generating callback token")
+			}
+			c.ARC.Callback.Token = callbackToken
+		}
+
+		options = append(options, engine.WithCallback(c.ARC.Callback.Host+BroadcastCallbackRoute, c.ARC.Callback.Token))
+	}
+	return options, nil
```
chris-4chain, line 297:
```suggestion
if !c.ARC.Callback.Enabled {
		return options, nil
	}
	if c.ARC.Callback.Token == "" {
		callbackToken, err := utils.HashAdler32(DefaultAdminXpub)
		if err != nil {
			return nil, spverrors.Wrapf(err, "error while generating callback token")
		}
		c.ARC.Callback.Token = callbackToken
	}

	options = append(options, engine.WithCallback(c.ARC.Callback.Host+BroadcastCallbackRoute, c.ARC.Callback.Token))
	return options, nil
```

config/config_to_options.go
```diff
@@ -0,0 +1,298 @@
...
+
+	var err error
+	if options, err = c.addClusterOpts(options); err != nil {
+		return nil, err
+	}
+
+	if options, err = c.addDataStoreOpts(options, testMode); err != nil {
+		return nil, err
+	}
+
```
chris-4chain, line 42: maybe the error-returning components could be at the top ?

config/config_to_options.go
```diff
@@ -0,0 +1,298 @@
...
+	if c.ExperimentalFeatures.PikePaymentEnabled {
+		options = append(options, engine.WithPaymailPikePaymentSupport())
+	}
+
+	return options
+}
+
+func (c *AppConfig) addTaskManagerOpts(options []engine.ClientOps) []engine.ClientOps {
+	// FIXME: Empty slice declaration using a literal
+	ops := []taskmanager.TasqOps{}
```
chris-4chain, line 251:
```suggestion
	// FIXME: Empty slice declaration using a literal
	var ops []taskmanager.TasqOps
```

config/config_to_options.go
```diff
@@ -0,0 +1,298 @@
...
+		return nil, err
+	}
+
+	options = c.addPaymailOpts(options)
+
+	options = c.addTaskManagerOpts(options)
+
+	options = c.addNotificationOpts(options)
+
+	options = c.addBroadcastClientOpts(options, logger)
```
chris-4chain, line 49: You forgot to apply `WithARC` which is the new way of ARC configuration for `ChainService`

config/config_to_options.go
```diff
@@ -0,0 +1,298 @@
...
+
+	if options, err = c.addCallbackOpts(options); err != nil {
+		return nil, err
+	}
+
+	options = c.addFeeQuotes(options)
+
+	return options, nil
```
chris-4chain, line 57: Also 
- `WithExcludedProviders`, 
- `WithCustomChainstate`, 
- `WithClusterClient`, 
- `WithPaymailServerConfig`, 
- `WithSQLConfigs`
...in `client_options.go` appear to be unused. 

I thing we can remove them - or consider to include.

config/load_test.go
chris-4chain, line 1: Consider renaming this `load_test.go` file - because it's kind of missleading, suggesting that it performs actual "Load Tests".

chris-4chain (changes requested):

logging/logging.go
dorzepowski: From what I remember, we're setting it to json because of kibana, that when it's starting as standard log the collector was failing because it was expecting a json.

config/config_to_options.go
```diff
@@ -0,0 +1,291 @@
...
+		return nil, err
+	}
+
+	options = c.addPaymailOpts(options)
+
+	options = c.addTaskManagerOpts(options)
+
+	options = c.addNotificationOpts(options)
+
+	options = c.addArcOpts(options)
```
chris-4chain, line 50:
```suggestion
	options = c.addARCOpts(options)
```

config/config_to_options.go
dorzepowski: Ahh I forgot to change it, and I put this fixme to make it later when I can run the app 😄

logging/logging.go
```diff
@@ -15,8 +16,22 @@ const (
...
+	if err != nil {
+		panic(err)
+	}
```
chris-4chain, line 24: Perhaps, instead of panic - we could return an error and handle it in `main`.
That would be consistent for `main`.

tests/tests.go
```diff
@@ -16,54 +19,71 @@ import (
 
 // TestSuite is for testing the entire package using real/mocked services
 type TestSuite struct {
-	AppConfig   *config.AppConfig   // App config
-	Router      *gin.Engine         // Gin router with handlers
-	Services    *config.AppServices // Services
-	suite.Suite                     // Extends the suite.Suite package
+	AppConfig *config.AppConfig // App config
+	Router    *gin.Engine       // Gin router with handlers
+	// Services    *config.AppServices // Services
```
chris-4chain, line 24: Or... 😄 just remove the commented-out `Services`.

logging/logging.go
dorzepowski: we're handling errors by logging it with use of logger,
so if in the first step we're not able to create a logger, it is better to panic,
beside that, this err shouldn't ever happen, because it can only happen if some programmer will use wrong (unknown) log level, so for me, such a mistakes should can be handled with panic.

config/config_to_options.go
dorzepowski: I will put it into separate task/pr
because we actually should check what are they doing and maybe remove the fields that they're setting.

chris-4chain (approved):

logging/logging.go
wregulski: Okay, it's clear for me now why it's different from defaults from config.example

wregulski (approved):

jakubmkowalski (approved):

PR #723, opened 2024-10-04
feat(SPV-1085) merkle roots verifier

chris-4chain:
### BHS Integration:

* [`config/check_block_header_service.go`](diffhunk://#diff-35080d3fdb816f3c864c63c0f8258689329e74d29aaaef81db251b639aeb00e2L10-R10): Updated imports and response models to use `chainmodels` instead of `chainstate` for BHS integration. [[1]](diffhunk://#diff-35080d3fdb816f3c864c63c0f8258689329e74d29aaaef81db251b639aeb00e2L10-R10) [[2]](diffhunk://#diff-35080d3fdb816f3c864c63c0f8258689329e74d29aaaef81db251b639aeb00e2L70-R77)
* [`config/services.go`](diffhunk://#diff-bd9e4264da754983c9dac3ee303a2f6a3a969a605b55d7cd1fcb8d74a43629afR134-R135): Added BHS configuration options to the `loadSPVWallet` function.
* [`engine/chain/chain_service.go`](diffhunk://#diff-31c2f462dcc0321b0c576079b4309bf2e50f720333016ad708804a719e2bf03aR5-R20): Added BHS service to the `chainService` struct and updated the `NewChainService` function to include BHS configuration.

### Error Handling:

* [`engine/chain/errors/bhs_errors.go`](diffhunk://#diff-f41f7189841c92ff4bfcb5dd99064b897264217468b972d9be74456399810968R1-R18): Introduced new error types specific to BHS, such as `ErrBHSUnreachable`, `ErrBHSNoSuccessResponse`, `ErrBHSUnauthorized`, `ErrBHSBadRequest`, and `ErrBHSUnhealthy`.

### Interface Updates:

* [`engine/chain/interface.go`](diffhunk://#diff-09b52f13f43dbaefbf98d9c12583aab9e3750313cc87b043f4e5e3401e351c15R16-R25): Added `BHSService` interface and included it in the main `Service` interface.

### Tests and Mocking:

* [`engine/chain/internal/bhs/bhs_mock_test.go`](diffhunk://#diff-707eebfb43a044b39e38af12040967b43e7e1aef34a21c8622500476f687a659R1-R52): Added mock tests for BHS verification.
* [`engine/chain/internal/bhs/healthcheck_test.go`](diffhunk://#diff-4b8e030ebfcee80273e9762c9614677116a2edadc285069e189ba208d4d0b80cR1-R51): Added tests for BHS health checks.

These changes ensure that the Block Header Service is properly integrated and that the system can handle various BHS-related errors effectively.

# Pull Request Checklist


engine/chain/internal/bhs/verify_mr.go
```diff
@@ -0,0 +1,55 @@
...
+	case http.StatusOK:
+		return result, nil
+	case http.StatusUnauthorized, http.StatusForbidden:
+		return nil, chainerrors.ErrBHSUnauthorized
+	case http.StatusNotFound:
+		return nil, chainerrors.ErrBHSUnreachable
+	case http.StatusBadRequest:
+		// Note: in case of error, BHS returns a string (not json) with the error message
+		// Most common error is "at least one merkleroot is required"
+		return nil, chainerrors.ErrBHSBadRequest.Wrap(spverrors.Newf("BHS response status code: %v", string(response.Body())))
```
dzolt-4chain, line 51: I think it should be something more like
`spverrors.Newf("BHS response status code: %d. Message: %v", response.StatusCode(), string(response.Body()))` or just change "status code" to "BHS error message: %v"

engine/chain/internal/bhs/heathcheck.go
```diff
@@ -0,0 +1,34 @@
...
+var pastMerkleRoot = []*spv.MerkleRootConfirmationRequestItem{{
+	// This is a known merkle root from the blockchain
+	// timestamp: 2024-10-04 10:55:37 UTC
+	MerkleRoot:  "f67ae53720205a55f4e99c632debabb68b6df0dc0f68affd200a076aee6e80e6",
+	BlockHeight: 864921,
+}}
+
+// HealthcheckBHS checks the health of the Block Headers Service (BHS).
+// It verifies the BHS is reachable and ready to verify merkle roots by providing a known merkle root.
+func (s *Service) HealthcheckBHS(ctx context.Context) error {
```
dzolt-4chain, line 22: Could this healthcheck be a simple GET to `/status` endpoint in BHS? I think they would actually check the same if the server response, and neither of them checks if the server is responsive i.e 1 second after the 1st request failed as mentioned in one of my PRs. It would simplify logic a little but maybe I'm missing something. WDYT?

engine/chain/internal/bhs/healthcheck_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/bitcoin-sv/spv-wallet/engine/tester"
+	"github.com/stretchr/testify/require"
+)
+
+/**
+NOTE: switch httpClient to resty.New() tu call actual ARC server
+*/
+
+func TestHealthcheckBHS(t *testing.T) {
+	t.Run("BHS Healthcheck success", func(t *testing.T) {
```
dzolt-4chain, line 19: Same here as my comment about healthcheck

engine/chain/internal/bhs/bhs_mock_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"time"
+
+	"github.com/bitcoin-sv/go-paymail/spv"
+	"github.com/bitcoin-sv/spv-wallet/engine/chain/models"
+	"github.com/go-resty/resty/v2"
+	"github.com/jarcoal/httpmock"
+)
+
+const (
+	bhsURL   = "http://localhost:8090"
```
dzolt-4chain, line 16: I think port should be 8080, but it's a mock so not sure if really an issue

engine/chain/internal/bhs/verify_mr.go
chris-4chain: Of course 👍 
Thx 😎

engine/chain/internal/bhs/heathcheck.go
chris-4chain: Nope, I guess 🙈 
Please check out how it is done right now in `check_block_header_service.go`.

We need to check if `BHS` can verify our transactions - that's why we need to have relatively updated/synced working instance of BHS - and this way we can check that. 

Maybe the name `HealthcheckBHS` is missleading, but:
- I wanted to reflect what we already have in  `check_block_header_service.go`
- I recall we discussed about it a few months ago and decided to do it that way. 

But maybe we want to change that logic 🤔 WDYT @dorzepowski @wregulski

engine/chain/internal/bhs/bhs_mock_test.go
chris-4chain: On my local machine I have `8090`.
Will change to default one 👍

engine/chain/internal/bhs/heathcheck.go
dorzepowski: I don't remember this discussion, but if we already discussed this, and agreed upon this solution, then lets keep it as is.

engine/chain/internal/bhs/verify_mr.go
```diff
@@ -0,0 +1,55 @@
...
+	switch response.StatusCode() {
+	case http.StatusOK:
+		return result, nil
+	case http.StatusUnauthorized, http.StatusForbidden:
+		return nil, chainerrors.ErrBHSUnauthorized
+	case http.StatusNotFound:
+		return nil, chainerrors.ErrBHSUnreachable
+	case http.StatusBadRequest:
+		// Note: in case of error, BHS returns a string (not json) with the error message
+		// Most common error is "at least one merkleroot is required"
```
dorzepowski, line 50: maybe we should check that the provided list is not empty before we do any call?

engine/paymail_service_provider.go
```diff
@@ -182,14 +182,24 @@ func (p *PaymailDefaultServiceProvider) VerifyMerkleRoots(
 		}()
 	}
 
-	request := make([]chainstate.MerkleRootConfirmationRequestItem, 0, len(merkleRoots))
-	for _, m := range merkleRoots {
-		request = append(request, chainstate.MerkleRootConfirmationRequestItem{
-			MerkleRoot:  m.MerkleRoot,
-			BlockHeight: m.BlockHeight,
-		})
+	confirmations, err := p.client.Chain().VerifyMerkleRoots(ctx, merkleRoots)
```
dorzepowski, line 185: IMHO previous approach was better,
I mean, that we verify merkle roots inside the Chainstate (currently Chain), and not returning the response here.
From the "client" of the VerifyMerkleRoots method perspective, all I want to is to either have information that it's valid or it's not valid - and then probably a proper detailed error could be returned.

engine/chain/internal/bhs/verify_mr_test.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/bitcoin-sv/go-paymail/spv"
+	"github.com/bitcoin-sv/spv-wallet/engine/chain"
+	"github.com/bitcoin-sv/spv-wallet/engine/chain/errors"
+	"github.com/bitcoin-sv/spv-wallet/engine/chain/models"
+	"github.com/bitcoin-sv/spv-wallet/engine/tester"
+	"github.com/stretchr/testify/require"
+)
+
+/**
+NOTE: switch httpClient to resty.New() tu call actual ARC server
```
dorzepowski, line 17: We don't make calls to ARC here 😄

dorzepowski (changes requested):

engine/chain/internal/bhs/heathcheck.go
```diff
@@ -0,0 +1,34 @@
...
+	timedCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	confirmations, err := s.VerifyMerkleRoots(timedCtx, pastMerkleRoot)
+	if err != nil {
+		return chainerrors.ErrBHSUnhealthy.Wrap(err)
+	}
+	if confirmations.ConfirmationState != chainmodels.MRConfirmed {
```
wregulski, line 30: I have same thoughts about this healtcheck as @chris-4chain but I am more concerned about this if clause - if we shouldn't handle all possible ConfirmationStates with proper description.

Like: What does it mean that we reached for example:
`MRUnableToVerify` that if it repeats there may be a problem with syncing with Tip of chain and it's worth to look at your chain tracker (BHS).

engine/chain/internal/bhs/verify_mr_test.go
chris-4chain: Changed to `BHS` 👍

engine/paymail_service_provider.go
chris-4chain: Let's discuss it on huddle 😄 - because I have some points on that.

engine/chain/internal/bhs/verify_mr.go
chris-4chain: I've added the empty-list checker at the beginning of this method. 👍 However I'd like to leave this `StatusBadRequest` case, because other error can be sent from BHS.

engine/chain/internal/bhs/heathcheck.go
chris-4chain: So let's keep it 👌

engine/chain/internal/bhs/healthcheck_test.go
chris-4chain: Resolved according to discussion above ☝️.

cmd/main.go
```diff
@@ -67,7 +67,16 @@ func main() {
 	}
 
 	// Try to ping the Block Headers Service if enabled
-	appConfig.CheckBlockHeadersService(appCtx, &logger)
+	if appConfig.IsBeefEnabled() {
```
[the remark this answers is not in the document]
chris-4chain: I've moved the logic, under `LogBHSReadiness` method.

wregulski (approved):

engine/paymail_service_provider.go
chris-4chain: I've adjusted it as we talked.
- Noe the VerifyMerkleRoots returns `(valid bool, err error)`

engine/chain/internal/bhs/verify_mr.go
```diff
@@ -0,0 +1,76 @@
...
+		s.logger.Warn().Msg("BHS is up but could not verify some merkle root(s). Defaulting to treat the provided merkle roots as valid.")
+		return true, nil
+	case chainmodels.MRInvalid:
+		return false, nil
+	default:
+		return false, spverrors.ErrInternal.Wrap(spverrors.Newf("unexpected confirmation state"))
+	}
+}
+
+func (s *Service) mrVerifyRequest(ctx context.Context, merkleRoots []*spv.MerkleRootConfirmationRequestItem) (*chainmodels.MerkleRootsConfirmations, error) {
```
dorzepowski, line 41: some more clear name please 🙏 
`requestForVerificationOfMerkleRoots`
`callToVerifyMerkleRoots`
`makeVerifyMerkleRootsRequest`

engine/chain/internal/bhs/verify_mr.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/bitcoin-sv/go-paymail/spv"
+	"github.com/bitcoin-sv/spv-wallet/engine/chain/errors"
+	"github.com/bitcoin-sv/spv-wallet/engine/chain/models"
+	"github.com/bitcoin-sv/spv-wallet/engine/spverrors"
+)
+
+/**
+NOTE: When you switch from enabled use_beef to disabled and restart the server,
+sometimes paymail capabilities are cached and VerifyMerkleRoots is called even though it shouldn't be.
+*/
```
dorzepowski, line 19: Do we need this note 🙈 
this service could be called in any other flow, not only paymail, and nobody will care to look if there is some comment with note.

engine/chain/models/mr_confirmation.go
```diff
@@ -0,0 +1,30 @@
...
+	BlockHeight  uint64                      `json:"blockHeight"`
+	MerkleRoot   string                      `json:"merkleRoot"`
+	Confirmation MerkleRootConfirmationState `json:"confirmation"`
+}
+
+// MerkleRootsConfirmations is an API response for confirming
+// merkle roots inclusion in the longest chain.
+type MerkleRootsConfirmations struct {
+	ConfirmationState MerkleRootConfirmationState `json:"confirmationState"`
+	Confirmations     []MerkleRootConfirmation    `json:"confirmations"`
```
dorzepowski, line 29: We don't use this part of the response, maybe it's good idea to just skip this field in a struct and not process it at all

engine/chain/internal/bhs/verify_mr_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+func TestVerifyMerkleRoots(t *testing.T) {
+	tests := map[string]struct {
+		request  []*spv.MerkleRootConfirmationRequestItem
+		response *chainmodels.MerkleRootsConfirmations
+		verified bool
+	}{
+		"Verify for CONFIRMED merkle roots": {
+			request: validMerkleRootsReq,
+			response: &chainmodels.MerkleRootsConfirmations{
```
dorzepowski, line 39: What if you would simply use here a json string?

engine/chain/internal/bhs/bhs_mock_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	responder := func(req *http.Request) (*http.Response, error) {
+		if applyTimeout {
+			time.Sleep(100 * time.Millisecond)
+		}
+		if req.Header.Get("Authorization") != "Bearer "+bhsToken {
+			return httpmock.NewStringResponse(http.StatusUnauthorized, ""), nil
+		}
+		var reqBody []*spv.MerkleRootConfirmationRequestItem
+		_ = json.NewDecoder(req.Body).Decode(&reqBody)
```
dorzepowski, line 33: if you would accept a string then you won't need to make a json from some struct, and you won't need to handle (or skip handling) an error in such case.

dorzepowski (changes requested):

engine/chain/internal/bhs/verify_mr.go
chris-4chain: Changed to `makeVerifyMerkleRootsRequest`.

engine/chain/internal/bhs/verify_mr.go
chris-4chain: We don't need it - but when I delete it, this knowledge will stay only in our heads :D 

But ok, removed 👍 I agree that this is not the best for it.

engine/chain/internal/bhs/bhs_mock_test.go
chris-4chain: Right 👍

dorzepowski (approved):

wregulski:
testing

PR #730, opened 2024-10-09
test(SPV-1086): get shared config endpoint tests

dorzepowski:
# Pull Request Checklist


actions/testabilities/assert_spvwallet_application.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (a *responseAssertions) assertJSONContentType() {
+	contentType := a.response.Header().Get("Content-Type")
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	a.require.NoError(err, "failed to parse content type")
+	a.assert.Equal("application/json", mediaType, "JSON content type expected on response")
+}
+
+func (a *responseAssertions) assertJSONBody(expectedFormat string, args []any) {
```
chris-4chain, line 75:
```suggestion
func (a *responseAssertions) assertJSONBody(expectedFormat string, args ...any) {
```

actions/testabilities/fixture_spvwallet_application.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (f *appFixture) NewTest(t testing.TB) SPVWalletApplicationFixture {
+	return newOf(*f, t)
+}
+
+func newOf(f appFixture, t testing.TB) *appFixture {
+	f.t = t
+	f.logger = tester.Logger(t)
+	return &f
+}
```
chris-4chain, line 75: The `newOf` is only used in `NewTest` function so maybe you could just move the code to the `NewTest` and skip the `newOf`?

actions/testabilities/fixture_spvwallet_application.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+	}
+}
+
+// initDbConnection creates a new connection that will be used as connection for engine
+func (f *appFixture) initDbConnection() {
+	hex, err := utils.RandomHex(8)
+	require.NoErrorf(f.t, err, "cannot generate random hex for sqlite inmemory db name")
+
+	f.dbConnectionString = "file:" + hex + "?mode=memory&cache=shared"
```
chris-4chain, line 147: Consider not using `cache=shared` to preserve each test cases separation.

actions/testabilities/ability_spvwallet_application.go
```diff
@@ -0,0 +1,11 @@
+package testabilities
+
+import "testing"
+
+func New(t testing.TB) (given SPVWalletApplicationFixture, then SPVWalletApplicationAssertions) {
+	return Given(t), Then(t)
+}
```
chris-4chain, line 7: The `New` function seems to be unused.

engine/tester/fixtures/users_fixtures.go
```diff
@@ -0,0 +1,117 @@
...
+func (f *User) XPubID() string {
+	xpub := f.XPub()
+	if xpub == "" {
+		return ""
+	}
+	return utils.Hash(xpub)
+}
+
+// All returns all fixtures.
+func All() []User {
```
chris-4chain, line 99: From the user-of-this-method perspective it is:
```
fixtures.All()
```
So, consider add to the name what it is behind the `All` 🙈 
```suggestion
func AllUsers() []User {
```
Also please adjust the desc/comment for this method.

actions/testabilities/fixture_spvwallet_application.go
dorzepowski: maybe you're right, maybe there is no good reason to use a cache=shared

actions/testabilities/ability_spvwallet_application.go
dorzepowski: yes, in this certain example. Because we're reusing the same spv-wallet for each call, 
but it will be common to start spv-wallet in each test separately to not have polluted db,
so I don't want to remove it just to restore it soon.

engine/tester/fixtures/users_fixtures.go
dorzepowski: good point

chris-4chain (approved):

PR #736, opened 2024-10-14
refactor(SPV-1079): move merkleroots to spv-wallet service

dzolt-4chain:
# Pull Request Checklist


Nazarii-4chain (approved):
LGTM

engine/chain/internal/bhs/get_merkleroots_from_bhs_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/bitcoin-sv/spv-wallet/engine/chain"
+	"github.com/bitcoin-sv/spv-wallet/engine/chain/errors"
+	"github.com/bitcoin-sv/spv-wallet/engine/chain/models"
+	"github.com/bitcoin-sv/spv-wallet/engine/tester"
+	"github.com/bitcoin-sv/spv-wallet/models"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetMerkleRootsFromBHSSuccess(t *testing.T) {
+	t.Run("Get MerkleRoots success", func(t *testing.T) {
```
dorzepowski, line 18: Because it is used only for kinda proxy or reading, I would prefer to add `integration`/`endpoint` tests instead unit tests. 

If you want to make this change in a separate PR. So it wouldn't block this PR.

engine/chain/errors/bhs_error_response.go
```diff
@@ -0,0 +1,57 @@
...
+
+	chainmodels "github.com/bitcoin-sv/spv-wallet/engine/chain/models"
+	"github.com/bitcoin-sv/spv-wallet/engine/spverrors"
+	"github.com/bitcoin-sv/spv-wallet/models"
+	"github.com/go-resty/resty/v2"
+)
+
+// MapBHSErrorResponseToSpverror is a method that will check what kind of response came back from
+// Block Header Service and map it to spverror and set it to context
+func MapBHSErrorResponseToSpverror(res *resty.Response) models.SPVError {
```
dorzepowski, line 16: Why this is public, does it need to be in package errors? for me it's rather a part of the logic of the service

engine/chain/models/bhs_error.go
```diff
@@ -0,0 +1,7 @@
+package chainmodels
+
+// BHSError is an error response that is returned from Block Header Service (BHS)
+type BHSError struct {
```
dorzepowski, line 4: is it expose anywhere by a service, maybe it could be private and in the bhs package?

actions/merkleroots/get.go
```diff
@@ -6,18 +6,30 @@ import (
...
+// @Param		lastEvaluatedKey query string false "last processed merkleroot in client's database"
+// @Success		200 {object} models.MerkleRootsBHSResponse "Paged response with Merkle Roots array in content "
+// @Failure		400	"Bad request - batchSize must be 0 or a positive integer"
+// @Failure		404	"Not found - No block with provided merkleroot was found"
+// @Failure		409	"Conflict - Provided merkleroot is not part of the longest chain"
+// @Failure 	500	"Internal Server Error - cannot create Block Header Service url. Please check your configuration"
+// @Failure 	500	"Internal Server Error - Block Header Service cannot be requested"
+// @Failure 	500	"Internal Server Error - Error while fetching Merkle Roots"
+// @Router  /api/v1/merkleroots [get]
+// @Security	x-auth-xpub
```
dorzepowski, line 27: you need to generate swagger docs

dorzepowski (changes requested):

engine/chain/errors/bhs_error_response.go
dzolt-4chain: I made it public so it can be imported in bhs service, and put it in error package, because it seemed like a good place to handle error responses.  It is indeed only used in a service so I will move it

engine/chain/models/bhs_error.go
dzolt-4chain: moved

engine/chain/errors/bhs_error_response.go
dzolt-4chain: moved

actions/merkleroots/get.go
dzolt-4chain: done

dorzepowski (approved):

engine/chain/interface.go
```diff
@@ -15,6 +17,7 @@ type ARCService interface {
 
 // BHSService for querying BHS server.
 type BHSService interface {
+	GetMerkleRootsFromBHS(ctx context.Context, query url.Values) (*models.MerkleRootsBHSResponse, error)
```
chris-4chain, line 20: I would not add the `FromBHS` suffix as other methods (within BHSSErvice or ARCService) don't have such suffix. 
So I vote for `GetMerkleRoots`.

engine/chain/internal/bhs/bhs_error.go
```diff
@@ -0,0 +1,7 @@
+package bhs
+
+// bhsError is an error response that is returned from Block Header Service (BHS)
+type bhsError struct {
```
chris-4chain, line 4: Private (unexported) structs don't require description/comments. 
If you really want to keep it, please make it more descriptive and less obvious.

engine/chain/internal/bhs/bhs_error_response.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// mapBHSErrorResponseToSpverror is a method that will check what kind of response came back from
+// Block Header Service and map it to spverror and set it to context
+func mapBHSErrorResponseToSpverror(res *resty.Response) models.SPVError {
+	var responseErr bhsError
+
+	if err := json.Unmarshal(res.Body(), &responseErr); err != nil {
+
+		// TODO: remove after SPV-1106 is done
```
chris-4chain, line 21: Please add appropriate task to jira to not forget about this `TODO`.
- or point out in `SPV-1106` description to do so 😄

engine/chain/internal/bhs/bhs_error_response.go
```diff
@@ -0,0 +1,57 @@
...
+
+	switch responseErr.Code {
+	case "ErrInvalidBatchSize":
+		return chainerrors.ErrInvalidBatchSize
+	case "ErrMerkleRootNotFound":
+		return chainerrors.ErrMerkleRootNotFound
+	case "ErrMerkleRootNotInLC":
+		return chainerrors.ErrMerkleRootNotInLongestChain
+	case "error-unauthorized":
+		return chainerrors.ErrBHSUnauthorized
```
chris-4chain, line 47: Maybe strings "enum" would be good here?

engine/chain/internal/bhs/get_merkleroots_from_bhs.go
```diff
@@ -0,0 +1,64 @@
...
+		req.SetAuthToken(s.bhsCfg.AuthToken)
+	} else {
+		s.logger.Warn().Msg("warning creating Block Headers Service url - auth token is not set. Some requests might not work")
+	}
+
+	var response models.MerkleRootsBHSResponse
+	res, err := req.
+		SetResult(&response).
+		SetQueryParamsFromValues(query).
+		Get(bhsURL.String())
```
chris-4chain, line 37: In the `chain` package I've used the convention in which the `Get/Post` method call is not mexed up with the `req` configuration (e.g. SetResult`).

After adjustments it would look like this:
```golang
        var response models.MerkleRootsBHSResponse
	req := s.httpClient.R().
		SetContext(ctx).
		SetHeader("Content-Type", "application/json").
		EnableTrace().
		SetResult(&response).
		SetQueryParamsFromValues(query)

	if s.bhsCfg.AuthToken != "" {
		req.SetAuthToken(s.bhsCfg.AuthToken)
	} else {
		s.logger.Warn().Msg("warning creating Block Headers Service url - auth token is not set. Some requests might not work")
	}

	
	res, err := req.Get(bhsURL.String())
```

engine/chain/internal/bhs/get_merkleroots_from_bhs.go
```diff
@@ -0,0 +1,64 @@
...
+	"syscall"
+
+	"github.com/bitcoin-sv/spv-wallet/engine/chain/errors"
+	"github.com/bitcoin-sv/spv-wallet/engine/spverrors"
+	"github.com/bitcoin-sv/spv-wallet/models"
+)
+
+// GetMerkleRootsFromBHS returns Merkle Roots from Block Header Service
+func (s *Service) GetMerkleRootsFromBHS(ctx context.Context, query url.Values) (*models.MerkleRootsBHSResponse, error) {
+	bhsURL, err := s.createBHSURL("/chain/merkleroot")
```
chris-4chain, line 17: You could adjust also `makeVerifyMerkleRootsRequest` to use the `createBHSURL` helper method.

engine/chain/internal/bhs/get_merkleroots_from_bhs.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	req := s.httpClient.R().
+		SetContext(ctx).
+		SetHeader("Content-Type", "application/json").
+		EnableTrace()
```
chris-4chain, line 25: I'm not convinced if we need the `EnableTrace` here. I didn't apply it for other `resty`-requsts in the `chain` package. 
hmmm 🤔 

Maybe we could rather use [client.EnableTrace](https://pkg.go.dev/github.com/go-resty/resty/v2#Client.EnableTrace) for the whole resty client - and decide whether or not to use it on the level where the client is created ?

I'm wondering how this influence to performance. Perhaps we should enable trace only when `logger level is set to debug` 🤔

engine/chain/internal/bhs/get_merkleroots_from_bhs.go
```diff
@@ -0,0 +1,64 @@
...
+
+	var response models.MerkleRootsBHSResponse
+	res, err := req.
+		SetResult(&response).
+		SetQueryParamsFromValues(query).
+		Get(bhsURL.String())
+	if err != nil {
+		if errors.Is(err, syscall.ECONNREFUSED) {
+			return nil, chainerrors.ErrBHSUnreachable.Wrap(err)
+		}
```
chris-4chain, line 41: I did similar thing in my implementations. But slightly different.
I think, we should pick one. (I don't know, for now, which one is better  😄 )

My checking of error:
```golang
if err != nil {
		var e net.Error
		if errors.As(err, &e) {
			return nil, chainerrors.ErrBHSUnreachable.Wrap(err)
		}
		return nil, spverrors.ErrInternal.Wrap(err)
	}
```

engine/chain/internal/bhs/get_merkleroots_from_bhs.go
```diff
@@ -0,0 +1,64 @@
...
+
+	return &response, nil
+}
+
+// createBHSURL parses Block Header Url from configuration and constructs a valid
+// endpoint with provided endpointPath variable
+func (s *Service) createBHSURL(endpointPath string) (*url.URL, error) {
+	if s.bhsCfg.URL == "" {
+		s.logger.Error().Msgf("create Block Header Service URL - url not configured")
+	}
```
chris-4chain, line 56: I didn't check the URL at this level. And I'm not convinced we have to do so 🤔 
Besides, we do the verification at `config` level:
https://github.com/bitcoin-sv/spv-wallet/blob/ce45722045af9f82c57d4d8935492b52d4dd0e17/config/validate_arc.go#L21-L23

chris-4chain (changes requested):

engine/chain/internal/bhs/bhs_error_response.go
dzolt-4chain: after discussion, removed the comment and left this handle error here, It might be useful when server returns 404 not foundd which is a string by default

engine/chain/internal/bhs/bhs_error_response.go
dzolt-4chain: These are only cases that will occur from bhs, I'd honestly leave it as is but if you think there are any benefits of using string enums here let me know and i might change it.

engine/chain/internal/bhs/get_merkleroots_from_bhs.go
dzolt-4chain: done

engine/chain/internal/bhs/get_merkleroots_from_bhs.go
dzolt-4chain: Yeah, you're right it should rather be set for the client and not per request. I will remove it from here and I suggest we discuss it if we even want it set or in a particular scenario as you mentioned

engine/chain/internal/bhs/get_merkleroots_from_bhs.go
dzolt-4chain: done

engine/chain/internal/bhs/get_merkleroots_from_bhs.go
dzolt-4chain: Hmm, my approach will handle more narrow cases like when server is unresponsive, while yours will also handle cases like timeouts and temporary network failures, so i incline more towards your approach. I will adjust mine

engine/chain/internal/bhs/get_merkleroots_from_bhs.go
dzolt-4chain: mm yeah, I copied what I created in a previous version but since we are checking it at the config level there is no need to do it twice, removed

engine/chain/internal/bhs/get_merkleroots_from_bhs_test.go
dzolt-4chain: okay, lets make it in a separate pr

engine/chain/internal/bhs/bhs_error_response.go
chris-4chain: It's ok 👍

chris-4chain (approved):

PR #737, opened 2024-10-14
feat(SPV-1075) new broadcaster

chris-4chain:
# Pull Request Checklist


engine/sdk_tx_getter.go
```diff
@@ -0,0 +1,51 @@
...
+	defer cancel()
+
+	var hexes []struct {
+		Hex string
+	}
+
+	err := db.
+		WithContext(queryIDsCtx).
+		Model(&Transaction{}).
+		Where("id IN (?)", slices.Collect(ids)).
```
ac4ch, line 34: how big the collection of ids you think it will be ? asking to avoid passing large number of IDs to the IN

engine/sdk_tx_getter.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func newSDKTxGetter(client *Client) *sdkTxGetter {
+	return &sdkTxGetter{client: client}
+}
+
+func (g *sdkTxGetter) GetTransactions(ctx context.Context, ids iter.Seq[string]) ([]*sdk.Transaction, error) {
+	db := g.client.Datastore().DB()
+
+	queryIDsCtx, cancel := context.WithTimeout(ctx, 20*time.Second)
```
ac4ch, line 24:
```suggestion
	queryIDsCtx, cancel := context.WithTimeout(ctx, 20*time.Second)
```
the context is good for preventing long-running queries, but you might want to consider making this configurable, or at lease to move it to constant var

engine/sdk_tx_getter.go
```diff
@@ -0,0 +1,51 @@
...
+	defer cancel()
+
```
ac4ch, line 26: i think you can introduce here early return if ids are not provided
```go
// Collect IDs
	idSlice := slices.Collect(ids)
	if len(idSlice) == 0 {
		return nil, nil // return early if no ids provided
	}
```

engine/sdk_tx_getter.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	transactions := make([]*sdk.Transaction, 0, len(hexes))
+	for _, record := range hexes {
+		tx, err := sdk.NewTransactionFromHex(record.Hex)
+		if err != nil {
+			return nil, spverrors.Wrapf(err, "Cannot parse transaction hex")
+		}
+		transactions = append(transactions, tx)
+	}
```
ac4ch, line 49:
```suggestion
	transactions := make([]*sdk.Transaction, 0, len(hexes))
	for _, record := range hexes {
		tx, err := sdk.NewTransactionFromHex(record.Hex)
		if err != nil {
			return nil, spverrors.Wrapf(err, "Cannot parse transaction hex")
		}
		transactions = append(transactions, tx)
	}
```
```go
transactions := make([]*sdk.Transaction, 0, len(hexes))
mu := sync.Mutex{}
wg := sync.WaitGroup{}

for _, record := range hexes {
    wg.Add(1)
    go func(hex string) {
        defer wg.Done()
        tx, err := sdk.NewTransactionFromHex(hex)
        if err != nil {
            // Handle error (could log or collect the error for later return)
            return
        }
        mu.Lock()
        transactions = append(transactions, tx)
        mu.Unlock()
    }(record.Hex)
}

wg.Wait()
```
something to consider

engine/chain/internal/combined_txs_getter.go
```diff
@@ -0,0 +1,56 @@
...
+	"iter"
+)
+
+// CombinedTxsGetter is a TransactionsGetter that combines multiple TransactionsGetters
+type CombinedTxsGetter struct {
+	txsGetters []chainmodels.TransactionsGetter
+}
+
+// NewCombinedTxsGetter creates a new CombinedTxsGetter
+func NewCombinedTxsGetter(txsGetters ...chainmodels.TransactionsGetter) *CombinedTxsGetter {
```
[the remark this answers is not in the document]
chris-4chain: - I'll put the `containsNilGetter` checking to the "constructor" (`NewCombinedTxsGetter`).
- For `len(txsGetters) == 0` - it should return empty slice - so  I'll add early return in `GetTransactions` method.

engine/sdk_tx_getter.go
chris-4chain: Not big. Usually several; sometimes several dozen.

I think it is already a boost, because in previous implementation there was `id = ? or id = ? or ...`. 
And also, previously, the whole object (all columns) was selected. Now only `hex`es are obtained (`SELECT hex from ...`)

engine/sdk_tx_getter.go
chris-4chain: Hmmm 🤔 
Regarding the size of `hexes` - which usually is no more than several elements, I would not introduce it 🤞 

So, for now, it seems like premature optimization which (considering (light but non-zero) overhead of goroutines and another sync mechanizm) might work even slower.

I suppose, in the future, we will have to review the whole transaction-flow (after current remaking) and try to boost/optimize things up.  However, I would start with locating bottlenecks and then optimizing.

engine/chain/internal/combined_txs_getter.go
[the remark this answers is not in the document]
chris-4chain: Done

engine/sdk_tx_getter.go
ac4ch: ok

engine/sdk_tx_getter.go
ac4ch: ok

wregulski (approved):

engine/chain/internal/arc/broadcast.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/bitcoin-sv/spv-wallet/engine/spverrors"
+	"github.com/go-resty/resty/v2"
+)
+
+// Custom ARC defined http status codes
+const (
+	StatusNotExtendedFormat             = 460
+	StatusFeeTooLow                     = 465
+	StatusCumulativeFeeValidationFailed = 473
+)
```
kuba-4chain, line 22: Some more statuses I'd propose to handle, if you want to separate them from the general error `ErrARCUnprocessable`:
- 460 - `ErrStatusTxFormat` - if the transaction is in RawTx format and Arc cannot extend it (already here as `StatusNotExtendedFormat`)
- 463 - `ErrStatusMalformed` - if there's a problem with decoding BEEF (also if there's too many SIGOPS in scripts or if the tx is over 100 bytes)
- 465 - `ErrStatusFees` - if the fee is too low (already here)
- 467 - `ErrStatusMinedAncestorsNotFound` - if BEEF is missing any mined parents
- 469 - `ErrStatusValidatingMerkleRoots` - if merkle root validation failed (if the blocks are older than Arc retention rate - currently 28 days - then this error will not occur and validation will pass)
- 473 - `ErrStatusCumulativeFees` - cumulative fee for all blocks is too low (I don't believe you need this, unless you're using CFV)

engine/chain/internal/arc/broadcast.go
```diff
@@ -0,0 +1,106 @@
...
+	switch response.StatusCode() {
+	case http.StatusOK:
+		if result.TXStatus.IsProblematic() {
+			return nil, chainerrors.ErrARCProblematicStatus.Wrap(spverrors.Newf("ARC Problematic tx status: %s", result.TXStatus))
+		}
+		return result, nil
+	case http.StatusUnauthorized, http.StatusForbidden, http.StatusNotFound:
+		return nil, s.wrapARCError(chainerrors.ErrARCUnauthorized, arcErr)
+	case http.StatusConflict:
+		return nil, s.wrapARCError(chainerrors.ErrARCGenericError, arcErr)
```
kuba-4chain, line 58: You can remove this, I believe Arc is never returning that :)

engine/client_options.go
```diff
@@ -579,46 +532,21 @@ func WithNotifications() ClientOps {
...
-		}
-		c.chainstate.options = append(c.chainstate.options, chainstate.WithCallback(callbackURL, callbackToken))
+		c.feeUnit = &feeUnit
 	}
 }
 
-// WithARC set ARC url params
-func WithARC(url, token, deploymentID string) ClientOps {
+// WithARC sets all the ARC options needed for broadcasting, querying transactions etc.
+func WithARC(arcCfg chainmodels.ARCConfig) ClientOps {
```
dorzepowski, line 547: are we making any validation of this config?

engine/chain/internal/combined_txs_getter.go
```diff
@@ -0,0 +1,71 @@
...
+
+	sdk "github.com/bitcoin-sv/go-sdk/transaction"
+	"github.com/bitcoin-sv/spv-wallet/engine/chain/errors"
+	"github.com/bitcoin-sv/spv-wallet/engine/chain/models"
+	"github.com/bitcoin-sv/spv-wallet/engine/spverrors"
+	"iter"
+)
+
+// CombinedTxsGetter is a TransactionsGetter that combines multiple TransactionsGetters
+type CombinedTxsGetter struct {
```
dorzepowski, line 15: Could it be private?

engine/model_transaction_config.go
```diff
@@ -29,7 +29,7 @@ type TransactionConfig struct {
 	ChangeSatoshis             uint64               `json:"change_satoshis" toml:"change_satoshis" yaml:"change_satoshis"` // The satoshis used for change
 	ExpiresIn                  time.Duration        `json:"expires_in" toml:"expires_in" yaml:"expires_in"`                // The expiration time for the draft and utxos
 	Fee                        uint64               `json:"fee" toml:"fee" yaml:"fee"`                                     // The fee used for the transaction (auto generated)
-	FeeUnit                    *bsv.FeeUnit         `json:"fee_unit" toml:"fee_unit" yaml:"fee_unit"`                      // Fee unit to use (overrides chainstate if set)
+	FeeUnit                    *bsv.FeeUnit         `json:"fee_unit" toml:"fee_unit" yaml:"fee_unit"`                      // Fee unit to use (overrides global FeeUnit if set)
```
dorzepowski, line 32:
```suggestion
	FeeUnit                    *bsv.FeeUnit         `json:"fee_unit" toml:"fee_unit" yaml:"fee_unit"`                      // Fee unit to use for this transaction (prevents usage of fee units configured at the server side)
```

engine/sdk_tx_getter.go
```diff
@@ -0,0 +1,59 @@
...
+
+type sdkTxGetter struct {
+	client *Client
+}
+
+func newSDKTxGetter(client *Client) *sdkTxGetter {
+	return &sdkTxGetter{client: client}
+}
+
+func (g *sdkTxGetter) GetTransactions(ctx context.Context, ids iter.Seq[string]) ([]*sdk.Transaction, error) {
```
dorzepowski, line 24: 🤔 I wonder if we could move it to separate package!?

engine/tx_broadcast.go
```diff
@@ -4,43 +4,31 @@ import (
...
-func broadcastTransaction(ctx context.Context, tx *Transaction) error {
-	client := tx.Client()
-	chainstateSrv := client.Chainstate()
+func broadcastTransaction(ctx context.Context, txModel *Transaction) error {
+	client := txModel.Client()
 
-	defer recoverAndLog(tx.Client().Logger())
+	defer recoverAndLog(txModel.Client().Logger())
 
 	unlock, err := newWriteLock(
```
dorzepowski, line 16: Minor, but 
Why do we actually need this lock here 🤔

engine/chain/internal/arc/broadcast.go
```diff
@@ -0,0 +1,106 @@
...
+func (s *Service) prepareTxHex(ctx context.Context, tx *sdk.Transaction) (string, error) {
+	if s.arcCfg.TxsGetter == nil {
+		efHex, err := tx.EFHex()
+		if err == nil {
+			return efHex, nil
+		}
+		s.logger.Warn().Msg("TransactionsGetter is not set, can't convert transaction to EFHex. Using raw transaction hex as a fallback.")
+		return tx.Hex(), nil
+	}
+	converter := ef.NewConverter(s.arcCfg.TxsGetter)
```
dorzepowski, line 84: ### minor
Maybe we could create (or not create) a new converter at the moment of creation of service and change the if above to check if there is converter or not?

engine/chain/internal/combined_txs_getter_test.go
```diff
@@ -0,0 +1,152 @@
...
+	tx3Hex = "01000000026d9a56036235bb5b5e39b04b6f188c74bc45189a9122f235cad5f0c4b668817d010000006a47304402205bad758ddad1816827d2f8d683d055c95652fa3e8902c0af1a2009b039e360350220152ac3f8417ef311ff61636fa91e99782a77e4d1d192c33fb2f1f245913b879f412103daf4a6e60ee877fa7e0639e0a4f416e5c80dfa05cd17762d77c62391a3322a52ffffffff08b11a9c1c6534f7d07d5367d4d7b073ca7803139122cbdc06f5e77463746310010000006a47304402204c8d367127e9d68b79c4d40bded75e773194ced7ab07345df966d5cbabca460102204d029f791d80b75d59fd2ba7b9e186e1d449e2e4e5c3d03a0f3e9f98b4bd132041210330278625061e9bb6104e959851e6698f2b96333cbd7262ee3c22410f2e5cf4a1ffffffff0314000000000000001976a9149df7244a4edade5f8d06ad392d568bc62218fe3988ac34120000000000001976a9144d4dec9ae2199860fc6ebe9cd446472f32f6ffd388acb5000000000000001976a914fd8154ad427ca5cce7209c121a29452add41e25d88ac00000000"
+)
+
+func TestCombinedTxGetter(t *testing.T) {
```
dorzepowski, line 39: I would rather test the whole solution at the level of `Chain().Broadcast`
But I'm not sure if it would be very difficult.

engine/chain/internal/junglebus/txs_getter_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"testing"
+	"time"
+
+	chainerrors "github.com/bitcoin-sv/spv-wallet/engine/chain/errors"
+	"github.com/bitcoin-sv/spv-wallet/engine/chain/internal/junglebus"
+	"github.com/bitcoin-sv/spv-wallet/engine/tester"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTransactionGetter(t *testing.T) {
```
dorzepowski, line 15: I would rather test the whole solution (with the combined getter, and optionally turned on or off junglebus fallback) at the level of `Chain().Broadcast`
But I'm not sure if it would be very difficult.

dorzepowski (changes requested):

engine/sdk_tx_getter.go
chris-4chain: Perhaps yes. But in some next step.

engine/tx_broadcast.go
chris-4chain: Don't know exactly. I believe we should not move it to out "new transaction-flow".
For now, I would not touch this 🤞

engine/client_options.go
chris-4chain: At config level, yes.
https://github.com/bitcoin-sv/spv-wallet/blob/c25c6f8c7bdfe832e7908502ba4887158a26b132/config/validate_arc.go#L16

engine/chain/internal/junglebus/txs_getter_test.go
chris-4chain: After internal conversation, we decided to remove the tests and move some of the cases to `arc/broadcast_test.go` to test it from "not internally exported" `Broadcast` method.

engine/chain/internal/arc/broadcast.go
chris-4chain: Actually errors `460`, `465`, `473` are currently handled separately.

The rest is handled by `default` switch-case, However, the arc-returned error message with details is wrapped and logged so no error-context is missing. 

For now, it's no use case for making it more granular. So I'll keep it as is 🥲

engine/chain/internal/arc/broadcast.go
chris-4chain: Removed.

engine/chain/chain_service.go
```diff
@@ -19,8 +21,29 @@ func NewChainService(logger zerolog.Logger, httpClient *resty.Client, arcCfg cha
...
 	return &chainService{
 		arc.NewARCService(logger.With().Str("chain", "arc").Logger(), httpClient, arcCfg),
 		bhs.NewBHSService(logger.With().Str("chain", "bhs").Logger(), httpClient, bhsConf),
 	}
 }
+
+func setTxGetter(logger zerolog.Logger, httpClient *resty.Client, arcCfg *chainmodels.ARCConfig) {
+	if !arcCfg.UseJunglebus {
+		return
+	}
```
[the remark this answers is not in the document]
chris-4chain: Done - but I also reorganized the `setTxGetter` func so check it out :D 
(renamed to `addJunglebusTxsGetter`

engine/chain/internal/arc/fee_unit.go
```diff
@@ -0,0 +1,20 @@
+package arc
+
+import (
+	"context"
+
+	"github.com/bitcoin-sv/spv-wallet/models/bsv"
+)
+
+// GetFeeUnit returns the current fee unit from the ARC policy.
+func (s *Service) GetFeeUnit(ctx context.Context) (*bsv.FeeUnit, error) {
```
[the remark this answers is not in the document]
chris-4chain: Done 👌.
It's hard to say... but you were right 🙈 :D

engine/chain/chain_service.go
```diff
@@ -19,8 +21,29 @@ func NewChainService(logger zerolog.Logger, httpClient *resty.Client, arcCfg cha
...
+	junglebusTxsGetter := junglebus.NewJunglebusService(logger.With().Str("service", "junglebus").Logger(), httpClient)
+	if arcCfg.TxsGetter != nil {
```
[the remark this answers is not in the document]
chris-4chain: Done xD

engine/chain/internal/combined_txs_getter.go
chris-4chain: it is private now. 
But I've also reorganized things here; partially as you suggested to use `"Combine"` word  xD - check it out 👀 👍

engine/chain/internal/arc/broadcast.go
chris-4chain: Now, I'm creating efConverter at the service creation stage.

config/validate_app_config.go
```diff
@@ -32,5 +34,9 @@ func (c *AppConfig) Validate() error {
 		return err
 	}
 
+	if c.CustomFeeUnit != nil && (c.CustomFeeUnit.Bytes <= 0 || c.CustomFeeUnit.Satoshis < 0) {
+		return spverrors.Newf("invalid custom fee unit - satoshis: %d; bytes: %d", c.CustomFeeUnit.Satoshis, c.CustomFeeUnit.Bytes)
+	}
+
```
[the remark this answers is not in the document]
chris-4chain: Done 👍

cmd/main.go
```diff
@@ -70,6 +70,11 @@ func main() {
 		spvWalletEngine.LogBHSReadiness(appCtx)
 	}
 
+	if err = spvWalletEngine.AskForFeeUnit(appCtx); err != nil {
```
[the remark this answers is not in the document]
chris-4chain: I've moved the logic to `NewClient` function. Removed from main. And added `UseCustomFeeUnit` for unit tests.

engine/client.go
```diff
@@ -413,3 +369,30 @@ func (c *Client) LogBHSReadiness(ctx context.Context) {
 		logger.Info().Msg("Block Headers Service is ready to verify transactions.")
 	}
 }
+
+// FeeUnit will return the fee unit used for transactions
+func (c *Client) FeeUnit() bsv.FeeUnit {
+	if c.options.feeUnit == nil {
+		c.Logger().Warn().Msg("Fee unit is not set (not by ARC policy nor by custom fee unit). Using fallback fee unit.")
```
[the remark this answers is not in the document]
chris-4chain: The fee unit is always set. So I've removed the "is-nil" condition.

engine/chain/internal/junglebus/txs_getter_test.go
chris-4chain: I've moved the junglebus mock requests to `ARC` internal package and removed the tests from `junglebus` package.
The junglebus is now tested by the way of testing ARC.

image

engine/chain/internal/combined_txs_getter.go
```diff
@@ -61,11 +61,20 @@ func (ctg *CombinedTxsGetter) GetTransactions(ctx context.Context, ids iter.Seq[
...
+
+func filterNilGetters(txsGetters ...chainmodels.TransactionsGetter) []chainmodels.TransactionsGetter {
+	return slices.Collect(
+		func(yield func(chainmodels.TransactionsGetter) bool) {
+			for _, v := range txsGetters {
+				if v != nil && !yield(v) {
+					return
+				}
+			}
+		},
```
dorzepowski, line 78: I'm not sure if this is much cleaner then:
```suggestion
	result := make([]chainmodels.TransactionsGetter, 0)
	for _, getter := range txsGetters {
		if getter == nil {
			result = append(result, getter)
		}
	}
	return result
```

it would be if there would be some function `slices.Filter(txsGetters, func(i, item) bool { .... })` but I cannot find such thing 😞

engine/chain/internal/combined_txs_getter.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/bitcoin-sv/spv-wallet/engine/spverrors"
+	"iter"
+)
+
+// CombineTxsGetters creates a new CombinedTxsGetter
+func CombineTxsGetters(txsGetters ...chainmodels.TransactionsGetter) chainmodels.TransactionsGetter {
+	if len(txsGetters) == 0 {
+		return &emptyGetter{}
+	}
+	if len(txsGetters) == 1 {
```
dorzepowski, line 20: this should be rather after filterNilGetters

engine/chain/internal/arc/service.go
```diff
@@ -1,23 +1,50 @@
...
+	return service
+}
+
+type noTxsGettersEFConverter struct{}
+
+func (n *noTxsGettersEFConverter) Convert(_ context.Context, tx *sdk.Transaction) (string, error) {
+	efHex, err := tx.EFHex()
+	if err != nil {
+		return "", chainerrors.ErrEFConversion.Wrap(err)
+	}
```
dorzepowski, line 48: you changed the logic and now it's not returning rawTx if EFHex will fail, or maybe I'm missing something 🤷

dorzepowski (changes requested):

engine/chain/internal/arc/mock_test.go
```diff
@@ -0,0 +1,253 @@
...
+	//	"GET",
+	//	fmt.Sprintf("https://junglebus.gorillapool.io/v1/transaction/get/%s", wrongTxID),
+	//	httpmock.NewStringResponder(http.StatusNotFound, `"encoding/hex: invalid byte: U+0077 'w'"`),
+	//)
+	//
+	//transport.RegisterResponder(
+	//	"GET",
+	//	fmt.Sprintf("https://junglebus.gorillapool.io/v1/transaction/get/%s", unknownTx),
+	//	httpmock.NewStringResponder(http.StatusNotFound, `"tx-not-found"`),
+	//)
```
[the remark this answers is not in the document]
chris-4chain: 👀

engine/chain/internal/combined_txs_getter.go
chris-4chain: I've changed it to 
```golang
return slices.DeleteFunc(txsGetters, func(getter chainmodels.TransactionsGetter) bool {
		return getter == nil
	})
```

engine/chain/internal/combined_txs_getter.go
chris-4chain: I don't get your point.

engine/chain/internal/arc/service.go
chris-4chain: You're missing something. 
Check that:
```golang
if arcCfg.TxsGetter == nil {
		logger.Warn().Msg("No transactions getter provided. Unsourced transactions will be broadcasted as raw hex.")
		service.efConverter = &noTxsGettersEFConverter{}
	} else {
		service.efConverter = ef.NewConverter(arcCfg.TxsGetter)
	}
```

```golang
func (s *Service) prepareTxHex(ctx context.Context, tx *sdk.Transaction) (string, error) {
	efHex, err := s.efConverter.Convert(ctx, tx)
	if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
		return "", spverrors.ErrCtxInterrupted.Wrap(err)
	}
	if err != nil {
		// Log level is set to Info because it can happen in standard flow when source transaction is not from our wallet (and Junglebus is disabled)
		s.logger.Info().Err(err).Msg("Could not convert transaction to EFHex. Using raw transaction hex as a fallback.")
		return tx.Hex(), nil
	}
	return efHex, nil
}
```

engine/chain/internal/combined_txs_getter.go
chris-4chain: Aaa ok - changed 👍

config/validate_customfee_test.go
```diff
@@ -0,0 +1,66 @@
...
+			customFee: &config.FeeUnitConfig{
+				Satoshis: 1,
+				Bytes:    -1,
+			},
+			shouldFail: true,
+		},
+	}
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			err := test.customFee.Validate()
```
dorzepowski, line 58: Can you do it like int the other tests for validate in this package 🙈 
IMHO you can quite simply copy example from validate_paymail_test.go
Ohhh and please split success and error test cases.

Why?
Because they are testing the validate at the AppConfig level not at some level below, so we know that it will be validated when called at the whole app config.
It's easier to understand or change that many tests if they all looks very similar.

engine/chain/internal/combined_txs_getter.go
```diff
@@ -68,13 +70,7 @@ func (ctg *emptyGetter) GetTransactions(_ context.Context, _ iter.Seq[string]) (
...
-	return slices.Collect(
-		func(yield func(chainmodels.TransactionsGetter) bool) {
-			for _, v := range txsGetters {
-				if v != nil && !yield(v) {
-					return
-				}
-			}
-		},
-	)
+	return slices.DeleteFunc(txsGetters, func(getter chainmodels.TransactionsGetter) bool {
```
dorzepowski, line 73: uuu nice finding ❤️

engine/chain/internal/arc/service.go
dorzepowski: ahhh ok, it was moved a little bit. 👍

dorzepowski (changes requested):

config/validate_customfee_test.go
chris-4chain: Ok Ok 😄 
You've introduced some cool method how to do config tests and I didn't noticed that 🙈 

I'm adjusting :D

dorzepowski (approved):

PR #772, opened 2024-11-15
test(SPV-1058): env variable to use file db or postgres instead of memory

dorzepowski:
# Pull Request Checklist


wregulski (approved):

chris-4chain (approved):

engine/testabilities/fixture_engine.go
```diff
@@ -110,11 +119,25 @@ func (f *engineFixture) ConfigForTests(opts ...ConfigOpts) *config.AppConfig {
 
 // initDbConnection creates a new connection that will be used as connection for engine
 func (f *engineFixture) initDbConnection() {
+	mode := os.Getenv("TEST_DB_MODE")
+	if mode == "postgres" {
+		f.config.Db.Datastore.Engine = datastore.PostgreSQL
```
dzolt-4chain, line 124: I know it might be a little more work but maybe it's worth to actually including here some way to connect different db? So It doesn't collide with what we have in our dev environments? WDYT?

I mean we could have here maybe something like a function to check if some ENV variable is set we could use this instead and if not default to postgres?

dorzepowski: As discussed if someone had some requirements for that, he could implement this in separate PR :)

PR #775, opened 2024-11-19
Feat(SPV-1008): new way of deleting paymails

dzolt-4chain:
# Pull Request Checklist


engine/datastore/where_mocks_test.go
```diff
@@ -47,7 +47,7 @@ func mockClient(engine Engine) (*Client, *gorm.DB) {
 	client, _ := clientInterface.(*Client)
 	client.options.engine = engine
 	dialector := mockDialector(engine)
-	gdb, _ := gorm.Open(dialector, &gorm.Config{})
+	gdb, _ := gorm.Open(dialector, &gorm.Config{TranslateError: true})
```
mgosek-4chain, line 50: That's the general comment not related strictly to the PR. I see that for some reason there is no error checking in funcs: _mockDialector, mockClient_. I would suggest creating a separate task and marking them as test helpers.

https://google.github.io/styleguide/go/best-practices#leave-testing-to-the-test-function

mgosek-4chain (approved):

engine/datastore/where_mocks_test.go
chris-4chain: I talked to @dorzepowski about that, and generally we don't see a need to introduce it. Especially now where there are many more important tasks to do.

engine/model_paymail_addresses.go
```diff
@@ -21,14 +22,17 @@ import (
 type PaymailAddress struct {
 	// Base model
 	Model
+	// Use gorm DeletedAt to handle soft deletes for paymail. We need to make it embedded so that GORM can soft delete
+	// the model. DeletedAt in Model even though is the same type as gorm.DeletedAt doesn't allow soft deletes by GORM.
```
chris-4chain, line 26: Please rewrite this comment.

engine/datastore/models.go
```diff
@@ -371,15 +370,12 @@ func (c *Client) Raw(query string) *gorm.DB {
 // checkResult will check for records or error
 func checkResult(result *gorm.DB) error {
 	if result.Error != nil {
-		if errors.Is(result.Error, gorm.ErrRecordNotFound) {
-			return ErrNoResults
-		}
```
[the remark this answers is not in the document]
dzolt-4chain: Yes, that's why I've changed the check in other places not to use the custom errors but gorms. With translate errors it's all handled for us and there is no need to return different error in this method. If the error will be ErrRecordNotFound it will be returned in return result.Error

engine/datastore/models.go
```diff
@@ -371,15 +370,12 @@ func (c *Client) Raw(query string) *gorm.DB {
 // checkResult will check for records or error
 func checkResult(result *gorm.DB) error {
```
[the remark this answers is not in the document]
dzolt-4chain: I've noticed that some functions depend on the rowsAffected like the cron that cleans draft transactions. But yes this needs to be investigated

engine/datastore/models.go
```diff
@@ -371,15 +370,12 @@ func (c *Client) Raw(query string) *gorm.DB {
...
-		if errors.Is(result.Error, gorm.ErrRecordNotFound) {
-			return ErrNoResults
-		}
 		return result.Error
 	}
 
 	// We should actually have some rows according to GORM
 	if result.RowsAffected == 0 {
-		return ErrNoResults
+		return gorm.ErrRecordNotFound
```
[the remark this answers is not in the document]
dzolt-4chain: We previously returned a custom error rather then gorms. Now that in this pr i've introduced translation of errors to gorm we need to return a gorm not found error. This will be checked in other places for that error

engine/datastore/models.go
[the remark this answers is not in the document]
chris-4chain: We'll do it in some of the next tasks 👍 Thanks :)

engine/model_paymail_addresses.go
dzolt-4chain: rewrote

engine/datastore/where_mocks_test.go
chris-4chain: @mgosek-4chain  About the `t.Helper`

I talked to @dorzepowski about that, and generally we don't see a need to introduce it. Especially now where there are many more important tasks to do.

chris-4chain (approved):

PR #778, opened 2024-11-21
feat(SPV-1058): database transaction inputs selector

dorzepowski:
# Pull Request Checklist

### DIFF
--- a/engine/database/testabilities/fixture_database.go
+++ b/engine/database/testabilities/fixture_database.go
@@ -1,0 +1,71 @@
+package testabilities
+
+import (
+	"testing"
+
+	"github.com/bitcoin-sv/spv-wallet/engine/database"
+	testengine "github.com/bitcoin-sv/spv-wallet/engine/testabilities"
+	"github.com/bitcoin-sv/spv-wallet/models/bsv"
+	"github.com/stretchr/testify/require"
+	"gorm.io/gorm"
+)
+
+type DatabaseFixture interface {
+	DB() DatabaseDataFixtures
+	GormDB() *gorm.DB
+}
+
+type DatabaseDataFixtures interface {
+	HasUTXO() UserUtxoFixture
+}
+
+type UserUtxoFixture interface {
+	// OwnedBySender ensures that the UTXO is owned by the sender (which is default behavior of this fixture).
+	OwnedBySender() UserUtxoFixture
+	// P2PKH ensures that the UTXO is with P2PKH locking script (which is default behavior of this fixture).
+	P2PKH() UserUtxoFixture
+	// WithSatoshis sets the satoshis value of the UTXO.
+	WithSatoshis(satoshis bsv.Satoshis) UserUtxoFixture
+
+	Storable[database.UserUtxos]
+}
+
+type Storable[Data any] interface {
+	// Stored ensures that the data is stored in the database.
+	Stored() *Data
+}
+
+type databaseFixture struct {
+	t                testing.TB
+	db               *gorm.DB
+	utxoEntriesIndex uint32
+}
+
+func Given(t testing.TB) (given DatabaseFixture, cleanup func()) {
+	engineWithConfign, cleanup := testengine.Given(t).Engine()
+
+	db := engineWithConfign.Engine.Datastore().DB()
+	fixture := &databaseFixture{
+		t:  t,
+		db: db,
+	}
+
+	// TODO: remove this when we will include UserUtxos in the production code
+	err := fixture.db.AutoMigrate(&database.UserUtxos{})
+	require.NoError(t, err)
+
+	return fixture, cleanup
+}
+
+func (f *databaseFixture) DB() DatabaseDataFixtures {
+	return f
+}
+
+func (f *databaseFixture) GormDB() *gorm.DB {
+	return f.db
+}
+
+func (f *databaseFixture) HasUTXO() UserUtxoFixture {
+	f.utxoEntriesIndex++
+	return newUtxoFixture(f.t, f.db, f.utxoEntriesIndex)
+}
--- a/engine/database/testabilities/user_utxo_fixture.go
+++ b/engine/database/testabilities/user_utxo_fixture.go
@@ -38,7 +38,7 @@ func newUtxoFixture(t testing.TB, db *gorm.DB, index uint32) *userUtxoFixture {
 }
 
 func txIDTemplated(index uint) string {
-	return fmt.Sprintf("b6ebbe78d93e1b81dd2c9c0c6cd67f9bdf832e9c2bb12a1d57f30c%010d", index)
+	return fmt.Sprintf("a%010de1b81dd2c9c0c6cd67f9bdf832e9c2bb12a1d57f30cb6ebbe78d9", index)
 }
 
 func (f *userUtxoFixture) OwnedBySender() UserUtxoFixture {
--- a/engine/database/testabilities/user_utxo_fixture.go
+++ b/engine/database/testabilities/user_utxo_fixture.go
@@ -1,0 +1,74 @@
+package testabilities
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/bitcoin-sv/spv-wallet/engine/database"
+	"github.com/bitcoin-sv/spv-wallet/engine/tester/fixtures"
+	"github.com/bitcoin-sv/spv-wallet/models/bsv"
+	"gorm.io/gorm"
+)
+
+var FirstCreatedAt = time.Date(2006, 02, 01, 15, 4, 5, 7, time.UTC)
+
+type userUtxoFixture struct {
+	db                           *gorm.DB
+	t                            testing.TB
+	index                        uint
+	xpubID                       string
+	txID                         string
+	vout                         uint32
+	satoshis                     bsv.Satoshis
+	unlockingScriptEstimatedSize uint64
+}
+
+func newUtxoFixture(t testing.TB, db *gorm.DB, index uint32) *userUtxoFixture {
+	return &userUtxoFixture{
+		t:                            t,
+		db:                           db,
+		index:                        uint(index),
+		xpubID:                       fixtures.Sender.XPubID(),
+		txID:                         txIDTemplated(uint(index)),
+		vout:                         index,
+		satoshis:                     1,
+		unlockingScriptEstimatedSize: 106,
+	}
+}
+
+func txIDTemplated(index uint) string {
+	return fmt.Sprintf("b6ebbe78d93e1b81dd2c9c0c6cd67f9bdf832e9c2bb12a1d57f30c%010d", index)
+}
+
+func (f *userUtxoFixture) OwnedBySender() UserUtxoFixture {
+	f.xpubID = fixtures.Sender.XPubID()
+	return f
+}
+
+func (f *userUtxoFixture) P2PKH() UserUtxoFixture {
+	f.unlockingScriptEstimatedSize = fixtures.EstimatedUnlockingScriptSizeForP2PKH
+	return f
+}
+
+func (f *userUtxoFixture) WithSatoshis(satoshis bsv.Satoshis) UserUtxoFixture {
+	f.satoshis = satoshis
+	return f
+}
+
+func (f *userUtxoFixture) Stored() *database.UserUtxos {
+	utxo := &database.UserUtxos{
+		XPubID:                       f.xpubID,
+		TxID:                         f.txID,
+		Vout:                         f.vout,
+		Satoshis:                     uint64(f.satoshis),
+		UnlockingScriptEstimatedSize: f.unlockingScriptEstimatedSize,
+		Bucket:                       "bsv",
+		CreatedAt:                    FirstCreatedAt.Add(time.Duration(f.index) * time.Second), //nolint:gosec // this is used for testing and it should be fine even in case of integer overflow.
+		TouchedAt:                    FirstCreatedAt.Add(time.Duration(24) * time.Hour),
+	}
+
+	f.db.Create(utxo)
+
+	return utxo
+}
--- a/engine/database/user_utxos.go
+++ b/engine/database/user_utxos.go
@@ -1,0 +1,15 @@
+package database
+
+import "time"
+
+// UserUtxos is a table holding user's Unspent Transaction Outputs (UTXOs).
+type UserUtxos struct {
+	XPubID                       string `gorm:"primaryKey;column:xpub_id;uniqueIndex:idx_window,sort:asc,priority:1"`
+	TxID                         string `gorm:"primaryKey;uniqueIndex:idx_window,sort:asc,priority:4"`
+	Vout                         uint32 `gorm:"primaryKey;uniqueIndex:idx_window,sort:asc,priority:5"`
+	Satoshis                     uint64
+	UnlockingScriptEstimatedSize uint64
+	Bucket                       string    `gorm:"check:chk_not_data_bucket,bucket <> 'data'"`
+	CreatedAt                    time.Time `gorm:"uniqueIndex:idx_window,sort:asc,priority:3"`
+	TouchedAt                    time.Time `gorm:"uniqueIndex:idx_window,sort:asc,priority:2"`
+}
--- a/engine/testabilities/fixture_engine.go
+++ b/engine/testabilities/fixture_engine.go
@@ -2,7 +2,6 @@ package testabilities
 
 import (
 	"context"
-	"database/sql"
 	"errors"
 	"os"
 	"testing"
@@ -43,7 +42,6 @@ type EngineFixture interface {
 type EngineWithConfig struct {
 	Config config.AppConfig
 	Engine engine.ClientInterface
-	DB     *sql.DB
 }
 
 type engineFixture struct {
@@ -52,7 +50,6 @@ type engineFixture struct {
 	t                  testing.TB
 	logger             zerolog.Logger
 	dbConnectionString string
-	dbConnection       *sql.DB
 	externalTransport  *httpmock.MockTransport
 	paymailClient      *paymailmock.PaymailClientMock
 }
@@ -82,7 +79,7 @@ func (f *engineFixture) Engine() (walletEngine EngineWithConfig, cleanup func())
 
 func (f *engineFixture) EngineWithConfiguration(opts ...ConfigOpts) (walletEngine EngineWithConfig, cleanup func()) {
 	f.config = f.ConfigForTests(opts...)
-	f.initDbConnection()
+	f.prepareDBConfigForTests()
 
 	options, err := f.config.ToEngineOptions(f.logger)
 	require.NoError(f.t, err)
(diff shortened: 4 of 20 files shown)


engine/database/testabilities/fixture_database.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/bitcoin-sv/spv-wallet/engine/database"
+	testengine "github.com/bitcoin-sv/spv-wallet/engine/testabilities"
+	"github.com/bitcoin-sv/spv-wallet/models/bsv"
+	"github.com/stretchr/testify/require"
+	"gorm.io/gorm"
+)
+
+type DatabaseFixture interface {
+	DB() DatabaseDataFixtures
+	GormDB() *gorm.DB
```
dzolt-4chain, line 15: What do you think about having the GormDB() inside DatabaseDataFixtures, maybe change it's name to DatabaseFixtures and inside there we could have an Instance() method and access the db like that? 

I think having two objects name similar like DB() and GormDB() can be a little confusing. WDYT?

engine/database/user_utxos.go
```diff
@@ -0,0 +1,15 @@
+package database
+
+import "time"
+
+// UserUtxos is a table holding user's Unspent Transaction Outputs (UTXOs).
+type UserUtxos struct {
+	XPubID                       string `gorm:"primaryKey;column:xpub_id;uniqueIndex:idx_window,sort:asc,priority:1"`
+	TxID                         string `gorm:"primaryKey;uniqueIndex:idx_window,sort:asc,priority:4"`
+	Vout                         uint32 `gorm:"primaryKey;uniqueIndex:idx_window,sort:asc,priority:5"`
+	Satoshis                     uint64
```
dzolt-4chain, line 10: Why don't we want to store `Satoshis` and `UnlockingScriptEstimatedSize` in the table? maybe move them as last variables if they are not "gorm related"

engine/tester/tgorm/gorm_db_for_printing_sql.go
```diff
@@ -0,0 +1,78 @@
...
+			TablePrefix:   "xapi_",
+			SingularTable: false,
+		},
+	}
+}
+
+type doNothingConnectionPool struct {
+}
+
+func (m *doNothingConnectionPool) PrepareContext(_ context.Context, _ string) (*sql.Stmt, error) {
```
dzolt-4chain, line 64: What's the use case for those functions?

engine/transaction/outlines/internal/inputs/testabilities/assertions_inputs_selector.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (a assertion) WithoutError(err error) SuccessfullySelectedInputsAssertions {
+	a.t.Helper()
+	a.require.NoError(err)
+	return a
+}
+
+func (a assertion) SelectedInputs(inputs []*database.UserUtxos) SelectedInputsAssertions {
+	a.t.Helper()
```
chris-4chain, line 51: So you decided to give `Helper` a chance ?
But now, we will have inconsistent test helpers (with and without `t.Helper` call)

engine/database/testabilities/user_utxo_fixture.go
```diff
@@ -0,0 +1,74 @@
...
+		txID:                         txIDTemplated(uint(index)),
+		vout:                         index,
+		satoshis:                     1,
+		unlockingScriptEstimatedSize: 106,
+	}
+}
+
+func txIDTemplated(index uint) string {
+	return fmt.Sprintf("b6ebbe78d93e1b81dd2c9c0c6cd67f9bdf832e9c2bb12a1d57f30c%010d", index)
+}
```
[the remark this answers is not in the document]
dorzepowski: maybe, for me it's kinda easier for tracking what is this tx, so maybe it's useful also for setting up grandparents,
I'm just thinking if this wouldn't be useful to start the id with those numbers instead of ending it 🤔

engine/database/testabilities/user_utxo_fixture.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (f *userUtxoFixture) Stored() *database.UserUtxos {
+	utxo := &database.UserUtxos{
+		XPubID:                       f.xpubID,
+		TxID:                         f.txID,
+		Vout:                         f.vout,
+		Satoshis:                     uint64(f.satoshis),
+		UnlockingScriptEstimatedSize: f.unlockingScriptEstimatedSize,
+		Bucket:                       "bsv",
+		CreatedAt:                    FirstCreatedAt.Add(time.Duration(f.index) * time.Second), //nolint:gosec // this is used for testing and it should be fine even in case of integer overflow.
```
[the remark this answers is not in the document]
dorzepowski: maybe let's consider this if there would be more places where we're putting such a comment.
I don't want to disable this check, just because of single case.

engine/database/user_utxos.go
```diff
@@ -0,0 +1,15 @@
+package database
+
+import "time"
+
+// UserUtxos is a table holding user's Unspent Transaction Outputs (UTXOs).
+type UserUtxos struct {
```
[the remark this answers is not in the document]
dorzepowski: yes this is another dimension of output, and when recording transaction and the output became the input, it should be removed (deleted) from this table (for all users - in case if we will handle multi ownership of the output)

engine/testabilities/fixture_engine.go
```diff
@@ -117,8 +113,13 @@ func (f *engineFixture) ConfigForTests(opts ...ConfigOpts) *config.AppConfig {
 	return configuration
 }
 
-// initDbConnection creates a new connection that will be used as connection for engine
-func (f *engineFixture) initDbConnection() {
+// prepareDBConfigForTests creates a new connection that will be used as connection for engine
+func (f *engineFixture) prepareDBConfigForTests() {
+	if f.config.Db.Datastore.Engine != datastore.SQLite {
+		panic("Other datastore engines are not supported in tests (yet)")
+	}
```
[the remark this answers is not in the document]
dorzepowski: It is check if the passed configuration doesn't want to use Postgres,
because we don't support postgres in tests yet (the `env` solution should be used only for debugging and development)

engine/transaction/outlines/internal/inputs/inputs_query_composer.go
```diff
@@ -0,0 +1,72 @@
...
+	return fmt.Sprintf("ceil((sum(unlocking_script_estimated_size) over (order by touched_at ASC, created_at ASC, tx_id ASC, vout ASC) + %d + %d) / cast(%d as float)) * %d as fee_with_change_output", c.txWithoutInputsSize, estimatedChangeOutputSize, c.feeUnit.Bytes, c.feeUnit.Satoshis)
+}
+
+func (c *inputsQueryComposer) feeCalculatedWithoutChangeOutput() string {
+	return fmt.Sprintf("ceil((sum(unlocking_script_estimated_size) over (order by touched_at ASC, created_at ASC, tx_id ASC, vout ASC) + %d) / cast(%d as float)) * %d as fee_no_change_output", c.txWithoutInputsSize, c.feeUnit.Bytes, c.feeUnit.Satoshis)
+}
+
+func (c *inputsQueryComposer) remainingValue() string {
+	return fmt.Sprintf("sum(satoshis) over (order by touched_at ASC, created_at ASC, tx_id ASC, vout ASC) - %d as remaining_value", c.outputsTotalValue)
+}
```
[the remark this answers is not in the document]
dorzepowski: I thought about it, but I'm not sure if it isn't less readable 🤷

engine/transaction/outlines/internal/inputs/inputs_query_composer.go
```diff
@@ -0,0 +1,72 @@
...
+	utxoTab := c.utxos(db)
+	utxoWithChange := c.addChangeValueCalculation(db, utxoTab)
+	utxoWithMinChange := c.searchForMinimalChangeValue(db, utxoWithChange)
+	selectedOutpoints := c.chooseInputsToCoverOutputsAndFeesAndHaveMinimalChange(db, utxoWithMinChange)
+
+	res := db.Model(&database.UserUtxos{}).Where("(tx_id, vout) in (?)", selectedOutpoints)
+	return res
+}
+
+func (c *inputsQueryComposer) utxos(db *gorm.DB) *gorm.DB {
```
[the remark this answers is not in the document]
dorzepowski: option 1. 

If we're processing some transaction and we know that there is some output that is spendable by given user, then we're adding this record also,
It actually should have foreign key to outputs and xpubids, and probably could be somehow located in Output struct in a gorm model to make the whole insert atomic, but let's leave this step for later PRs

engine/transaction/outlines/internal/inputs/selector.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	return composer.build(db)
+}
+
+func (r *sqlInputsSelector) buildUpdateTouchedAtQuery(db *gorm.DB, utxos []*database.UserUtxos) *gorm.DB {
+	outpoints := make([][]any, 0, len(utxos))
+	for _, utxo := range utxos {
+		outpoints = append(outpoints, []any{utxo.TxID, utxo.Vout})
+	}
+	return db.Model(&database.UserUtxos{}).Where("(tx_id, vout) in (?)", outpoints)
```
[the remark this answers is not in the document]
dorzepowski: 🤦

engine/transaction/outlines/internal/inputs/testabilities/assertions_inputs_selector.go
dorzepowski: meh,
Actually I don't see any difference, so for me it's a minor thing to keep consistency in this matter.

engine/tester/tgorm/gorm_db_for_printing_sql.go
dorzepowski: As additional context, we use this in 2 places:
1. the tests for datastore custom `where clause`  - which is inheritance from BUX and we introduce those tests to ensure that actually generated SQL is SQLInjection proof (yes, because it wasn't previously)
2. I use it in examples for selector to show the generated sql query, because the query isn't trivial, and building it with gorm doesn't actually make it more readable, therefore thanks to testable examples, one can just copy the SQL from `// Output: ` and use it against the chosen database to check the result of the query, and eventually format the SQL string so that it would be more readable for him.

engine/database/testabilities/fixture_database.go
```diff
@@ -0,0 +1,71 @@
...
+	engineWithConfign, cleanup := testengine.Given(t).Engine()
+
+	db := engineWithConfign.Engine.Datastore().DB()
+	fixture := &databaseFixture{
+		t:  t,
+		db: db,
+	}
+
+	// TODO: remove this when we will include UserUtxos in the production code
+	err := fixture.db.AutoMigrate(&database.UserUtxos{})
```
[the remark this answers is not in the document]
dorzepowski: For now it shouldn't be a problem, because I'm not binding the model with the production code,
but later we would for sure need to discuss how to connect (connect not merge) the models and when to fill it and when to remove it and what I need to return from db exactly.

engine/testabilities/fixture_engine.go
```diff
@@ -117,8 +113,13 @@ func (f *engineFixture) ConfigForTests(opts ...ConfigOpts) *config.AppConfig {
 	return configuration
 }
 
-// initDbConnection creates a new connection that will be used as connection for engine
-func (f *engineFixture) initDbConnection() {
+// prepareDBConfigForTests creates a new connection that will be used as connection for engine
+func (f *engineFixture) prepareDBConfigForTests() {
+	if f.config.Db.Datastore.Engine != datastore.SQLite {
+		panic("Other datastore engines are not supported in tests (yet)")
```
mgosek-4chain, line 119: I strongly recommend using `t.Fatal` or `t.Fatal`f from the built-in `testing` package. These functions signal test failures in a controlled manner, ensuring the framework handles them internally (e.g., logging the current state and terminating the test). Unlike calling panic directly, they avoid unintended side effects, such as disrupting other tests or goroutines, providing a cleaner and more predictable testing process.
 
 
```
if f.config.Db.Datastore.Engine != datastore.SQLite {
		f.t.Fatal("Other datastore engines are not supported in tests (yet)")
	}
```

engine/tester/tgorm/gorm_db_for_printing_sql.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+// GormDBForPrintingSQL is creating a gorm.DB instance that can be used to print or check SQL (without connecting to postgres for example).
+// This is useful for checking generated SQL statements.
+func GormDBForPrintingSQL(dbType DBType) *gorm.DB {
+	dialector := chooseDialector(dbType)
+
+	db, err := gorm.Open(dialector, gormConfig())
+	if err != nil {
+		panic(err)
```
mgosek-4chain, line 31: I would recommend applying the suggestion from the previous comment. Additionally, we can make this function as test helper what in general can improve the failure reporting. 

https://google.github.io/styleguide/go/best-practices#error-handling-in-test-helpers

engine/transaction/outlines/internal/inputs/testabilities/assertions_inputs_selector.go
mgosek-4chain: Basically, there is a difference.  Take a look at the following example:

```
import "testing"

func FuncMarkedAsTestHelper(t *testing.T) {
	t.Helper()
	t.Fatal("something bad happend") // line 7
}

func FuncNotMarkedAsTestHelper(t *testing.T) {
	t.Fatal("something bad happend") // line 11
}

func Test(t *testing.T) {
	FuncNotMarkedAsTestHelper(t) // line 15
}
```

**Test output running FuncNotMarkedAsTestHelper call:** 
--- FAIL: Test (0.00s)    .../dummy_test.go:**11**: something bad happend
 
**Test output running FuncNotMarkedAsTestHelper call:** 
--- FAIL: Test (0.00s)    .../dummy_test.go:**15**: something bad happend

You can see the line reference is different.  Marking functions as `t.Helper()` helps in navigation and debugging by ensuring that when a test fails (due to `t.Fatal`, `t.Fatalf`, or a `panic`), the failure report points to the line in the test where the helper function was called, rather than the line inside the helper function itself.

engine/transaction/outlines/internal/inputs/selector_test.go
```diff
@@ -0,0 +1,158 @@
...
+			// given:
+			given, then, cleanup := testabilities.New(t)
+			defer cleanup()
+
+			// and: having some utxo in database
+			var ownedInputs []*database.UserUtxos
+			ownedInputs = append(ownedInputs, given.DB().HasUTXO().OwnedBySender().P2PKH().WithSatoshis(10).Stored())
+			ownedInputs = append(ownedInputs, given.DB().HasUTXO().OwnedBySender().P2PKH().WithSatoshis(10).Stored())
+			ownedInputs = append(ownedInputs, given.DB().HasUTXO().OwnedBySender().P2PKH().WithSatoshis(10).Stored())
+			ownedInputs = append(ownedInputs, given.DB().HasUTXO().OwnedBySender().P2PKH().WithSatoshis(10).Stored())
```
mgosek-4chain, line 138:
```
ownedInputs := []*database.UserUtxos{
		given.DB().HasUTXO().OwnedBySender().P2PKH().WithSatoshis(10).Stored(),
		given.DB().HasUTXO().OwnedBySender().P2PKH().WithSatoshis(10).Stored(),
		given.DB().HasUTXO().OwnedBySender().P2PKH().WithSatoshis(10).Stored(),
		given.DB().HasUTXO().OwnedBySender().P2PKH().WithSatoshis(10).Stored(),
}
```

engine/transaction/outlines/internal/inputs/selector.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/bitcoin-sv/spv-wallet/engine/spverrors"
+	txerrors "github.com/bitcoin-sv/spv-wallet/engine/transaction/errors"
+	"github.com/bitcoin-sv/spv-wallet/models/bsv"
+	"gorm.io/gorm"
+)
+
+const txIdColumn = "tx_id"
+const voutColumn = "vout"
+
+var model = &database.UserUtxos{}
```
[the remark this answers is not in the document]
mgosek-4chain: If it's not used anywhere that maybe we can remove it?  :)

engine/database/testabilities/user_utxo_fixture.go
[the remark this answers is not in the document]
dorzepowski: Shoul I change the pattern here, or we can leave it as is for now?

engine/transaction/outlines/internal/inputs/selector.go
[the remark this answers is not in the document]
dorzepowski: I'm removing it. It doesn't provide any benefits

engine/testabilities/fixture_engine.go
dorzepowski: I'm changing it to require.Equal

engine/transaction/outlines/internal/inputs/selector_test.go
dorzepowski: good point, I was thinking about other approach and forgot to make it better.

engine/transaction/outlines/internal/inputs/testabilities/assertions_inputs_selector.go
dorzepowski: 1. I know what is the difference, but when using testify, you're getting the whole stack and the file name is a littlebit secondary
2. I'm not sure if we we should somehow distiguish the usage of the t.Helper between env arrangement (when the fixtures are preapred and everything is established and the assertions. Because if something fails durring setup I would rather want to search for the place where it really fail, not the code that call that place, while when making assertions I'm rather interested in what were the values that wasn't expected and at what point in the test exectution it happens.

if you want to push for t.Helpers, I would rather move this discussion out of this PR, and agree on the standard within whole team and adjust existing code in some PR.

engine/database/testabilities/fixture_database.go
dzolt-4chain: It confused me when I was reading the code, but now that I tried to write a simple implementation of what I had in mind it's not a good idea to do it like this. I mixed up what the `DatabaseDataFixtures` actually do with what I thought it should do. We can discuss on priv but it was my mistake. So please ignore this comment or maybe just consider changing `GormDB()` to `DBInstance()` or `Instance()`. But if you think it doesn't give much value it's still fine.

engine/database/testabilities/user_utxo_fixture.go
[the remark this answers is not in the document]
chris-4chain: If it's quick - I will do it now. Otherwise - in the next occasion.

engine/transaction/outlines/internal/inputs/testabilities/assertions_inputs_selector.go
chris-4chain: I would resolve this conversation 🙈

chris-4chain (approved):

mgosek-4chain (approved):

PR #795, opened 2024-12-03
feat(SPV-1225): admin confirm contacts

jakubmkowalski:
This pull request introduces a new feature for confirming contacts in the admin panel, along with the necessary documentation and model updates. The most important changes include adding new endpoints for confirming contacts, updating the routes, and enhancing the documentation to reflect these changes.

### New Feature: Confirm Contacts

* [`actions/admin/contact.go`](diffhunk://#diff-5ead85ba379022ea0d41ca2ccd2bbb17264ab36f5fab34b78f477536241306adR214-R249): Added a new function `contactsConfirm` to handle the confirmation of contacts. This function processes incoming requests, maps the data, and interacts with the engine to confirm contacts.
* [`actions/admin/contact_old.go`](diffhunk://#diff-a525e17305da2496243f1d61cf844c4ca9b13b3759b303224daf9b4703a022cdR217-R252): Added a similar function `contactsConfirmOld` for the old version of the contacts API.
* [`actions/admin/routes.go`](diffhunk://#diff-2389fc4fa1cd88efd660d2a756bb24b5f2da971c4e55ab09f29cf05b811be5fdR20): Updated the routes to include the new POST endpoints for confirming contacts in both the current and old admin groups. [[1]](diffhunk://#diff-2389fc4fa1cd88efd660d2a756bb24b5f2da971c4e55ab09f29cf05b811be5fdR20) [[2]](diffhunk://#diff-2389fc4fa1cd88efd660d2a756bb24b5f2da971c4e55ab09f29cf05b811be5fdR54)

# Pull Request Checklist


actions/contacts/models.go
```diff
@@ -15,6 +15,16 @@ type UpsertContact struct {
...
 
+// ContactData is a type for contact data
+type ContactData struct {
+	// OwnerXpubID is a xpub id related to contact.
+	OwnerXpubID string `json:"xpub_id" toml:"xpub_id" yaml:"xpub_id" gorm:"column:xpub_id;<-:create;type:char(64);foreignKey:XpubID;reference:ID;index;comment:This is the related xPub"`
+	// FullName is name which could be shown instead of whole paymail address.
+	FullName string `json:"full_name" toml:"full_name" yaml:"full_name" gorm:"<-create;comment:This is the contact's full name"`
+	// Paymail is a paymail address related to contact.
+	Paymail string `json:"paymail" toml:"paymail" yaml:"paymail" gorm:"<-create;comment:This is the paymail address alias@domain.com"`
+}
```
chris-4chain, line 26: Why so many tag?
I'm 99% sure that even for `gorm` the `toml & yaml` taggs are not necessary.
But here - I don't think the `gorm` taggs are expected either - because I suppose this is not DB model.

Or I'm missing something 😄 🤔

engine/action_contact.go
```diff
@@ -403,6 +403,48 @@ func (c *Client) upsertContact(ctx context.Context, paymailService paymailclient
 	return contact, nil
 }
 
+// AdminConfirmContacts confirms provided contacts, should be used only by the admin.
+func (c *Client) AdminConfirmContacts(ctx context.Context, contacts []*Contact) error {
+	contactList := []*Contact{}
```
chris-4chain, line 408: Please preallocate memory for this slice - because we already know the final size.

chris-4chain (changes requested):

actions/contacts/models.go
jakubmkowalski: Only json tags are necessary. Rest of them removed.

engine/action_contact.go
jakubmkowalski: Size preallocated

actions/contacts/models.go
```diff
@@ -15,6 +15,16 @@ type UpsertContact struct {
 	RequesterPaymail string `json:"requesterPaymail"`
 }
 
+// ContactData is a type for contact data
+type ContactData struct {
+	// OwnerXpubID is a xpub id related to contact.
+	OwnerXpubID string `json:"xpub_id"`
```
dzolt-4chain, line 21: I think in new admin api we moved from snake case to camel case? I might be wrong but shouldn't there be xpubId and fullName or something in json struct?

dzolt-4chain (approved):

actions/contacts/models.go
jakubmkowalski: This structure is unused, left over from another approach to the solution. Deleted entirety

engine/action_contact.go
```diff
@@ -403,6 +403,48 @@ func (c *Client) upsertContact(ctx context.Context, paymailService paymailclient
...
+			c.logContactError(contact.OwnerXpubID, contact.Paymail, fmt.Sprintf("error while getting contact: %s", err.Error()))
+			return spverrors.Wrapf(err, "error while getting contact")
+		}
+
+		if cont == nil {
+			return spverrors.ErrContactsNotFound
+		}
+
+		contactList = append(contactList, cont)
+	}
```
mgosek-4chain, line 421: If I understand correctly, this approach would result in N queries to the database, where N is the number of contacts. If that's the case, I’d suggest considering a bulk query to retrieve the data with the minimal number of calls. Do we currently support this kind of functionality in our database implementation? 🤔

engine/action_contact.go
```diff
@@ -403,6 +403,48 @@ func (c *Client) upsertContact(ctx context.Context, paymailService paymailclient
...
+		err := tx.Commit()
+		if err != nil {
+			return _closeTxWithError(tx, err)
+		}
+
+		return nil
+	})
+
+	if err != nil {
+		c.logContactError("", "", fmt.Sprintf("error while confirming contacts: %s", err.Error()))
```
mgosek-4chain, line 441: Could you clarify the reason for skipping these two argument values? 🤔

actions/contacts/models.go
chris-4chain: 😄

engine/action_contact.go
chris-4chain: I wouldn't do this right now - because we want to make a transition from custom ("buggy") `dataStore` package to a approach leveraging GORM.
So, for now, I would suggest not to touch the `datasStore` package specially I wouldn't add any new features to it.

engine/action_contact.go
chris-4chain: Generally I wouldn't use `logContactError` at all.
I would rather wrap an error (one line below it) - and the error would be logged (unfolded) at the `endpoint` level and the topmost one returned to the client. 

But, now you only add only one function to the existing solution - so I would leave it as is.

actions/admin/contact.go
```diff
@@ -210,3 +211,39 @@ func contactsAccept(c *gin.Context, _ *reqctx.AdminContext) {
...
+// @Success		200
+// @Failure		400	"Bad request - Error while getting data from request body"
+// @Failure		404	"Not found - Error, contacts not found"
+// @Failure 	500	"Internal server error - Error, confirming contact failed"
+// @Router		/api/v1/admin/contacts/confirmations [post]
+// @Security	x-auth-xpub
+func contactsConfirm(c *gin.Context, _ *reqctx.AdminContext) {
+	logger := reqctx.Logger(c)
+
+	var reqParams []*models.ContactConfirmationData
```
dorzepowski, line 231: I would rather limit the number of contacts that could be confirmed at once to 2 contacts.
Maybe having object (instead of list) with fields `partyA` and `partyB` would be then more intuitive that you can only confirm 2 counterparties.
Also this should rather confirm the contacts against each other, not confirm any contact. So given party A and party B we want A to have B confirmed and B to have A confirmed, no other contacts.

BUT WHY?
Because confirmation is very important operation that is assuring the you will make a transfers to correct person/address, and unfortunately server is not able to prove that the counterparties really made this confirmation, and what's more, admin is unable to prove to server that they do that, so it's important to limit this operation scope to a bare minimum.

Also this would kinda solve the problem o N queries because N is always 2  😄 


And yes, I know:
me
(🇬🇧 Sorry, AFAIK, there is not English substitute for that )

actions/admin/contact_old.go
```diff
@@ -214,3 +214,39 @@ func contactsAcceptOld(c *gin.Context, _ *reqctx.AdminContext) {
...
+// @Tags		Admin
+// @Produce		json
+// @Param		[]models.ContactConfirmationData body []models.ContactConfirmationData true "Contacts data"
+// @Success		200
+// @Failure		400	"Bad request - Error while getting data from request body"
+// @Failure		404	"Not found - Error, contacts not found"
+// @Failure 	500	"Internal server error - Error, confirming contact failed"
+// @Router		/v1/admin/contacts/confirmations [post]
+// @Security	x-auth-xpub
+func contactsConfirmOld(c *gin.Context, _ *reqctx.AdminContext) {
```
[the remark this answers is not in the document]
jakubmkowalski: We didn't switch admin endpoints to use `v1/api/` endpoints so until we change that in clients we need also "old" admin endpoints

engine/action_contact.go
jakubmkowalski: Implementation based on earlier methods on contacts. In this context we don't know which xpub and paymail failed, so they are empty.

engine/action_contact.go
jakubmkowalski: Since there is currently a limit of 2 contacts to confirm in a request, the current implementation is not as much of a pain. As Chris mentioned, adding features to the `dataStore` is a waste of time.

mgosek-4chain (approved):

actions/admin/contact.go
chris-4chain: Hmmm... instead of array with a "not-explicit-from-user-perspective" limit.
Maybe: a model:
```goalng
type AdminConfirmContactPair struct {
    A ContactConfirmationData
    B ContactConfirmationData
}
```

chris-4chain (approved):

actions/admin/contact.go
jakubmkowalski: array swapped to:
```
type AdminConfirmContactPair struct {
	ContactA ContactConfirmationData `json:"contactA"`
	ContactB ContactConfirmationData `json:"contactB"`
}
```

dorzepowski (changes requested):

actions/admin/contact_old.go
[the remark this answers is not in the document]
jakubmkowalski: removed "old" one swagger docs, swap function for path

actions/admin/contact.go
```diff
@@ -210,3 +211,39 @@ func contactsAccept(c *gin.Context, _ *reqctx.AdminContext) {
 
 	c.JSON(http.StatusOK, contract)
 }
+
+// contactsConfirm will perform Confirm action on contacts with the given xpub ids and paymails
+// Perform confirm action on contacts godoc
+// @Summary		Confirm contacts
```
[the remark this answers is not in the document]
jakubmkowalski: changed

actions/admin/contact.go
```diff
@@ -210,3 +211,39 @@ func contactsAccept(c *gin.Context, _ *reqctx.AdminContext) {
...
+// @Description Confirm contacts
```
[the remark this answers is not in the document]
jakubmkowalski: changed

models/contact.go
```diff
@@ -25,6 +25,20 @@ type Contact struct {
 	Status response.ContactStatus `json:"status" example:"unconfirmed"`
 }
 
+type ContactConfirmationData struct {
+	// XPubID is a xpub id related to contact.
+	XPubID string `json:"xpubId" example:"68af358bde7d8641621c7dd3de1a276c9a62cfa9e2d0740494519f1ba61e2f4a"`
+	// FullName is name which could be shown instead of whole paymail address.
+	FullName string `json:"fullName" example:"Test User"`
```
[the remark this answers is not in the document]
jakubmkowalski: removed, as name validation will be performed when adding a contact

dorzepowski (approved):

models/contact.go
```diff
@@ -25,6 +25,18 @@ type Contact struct {
...
 
+type ContactConfirmationData struct {
+	// XPubID is a xpub id related to contact.
+	XPubID string `json:"xpubId" example:"68af358bde7d8641621c7dd3de1a276c9a62cfa9e2d0740494519f1ba61e2f4a"`
+	// Paymail is a paymail address related to contact.
+	Paymail string `json:"paymail" example:"test@test.test"`
+}
+
+type AdminConfirmContactPair struct {
+	ContactA ContactConfirmationData `json:"contactA"`
```
dorzepowski, line 36: As we talked,
maybe it would be much better to provide only the pair of paymail addresses. And the endpoint would check if ContacA has ContactB and if ContactB has ContactA and then and only then it will confirm this pair.

dorzepowski (changes requested):

dorzepowski:
Overall very good job

actions/admin/contact.go
```diff
@@ -210,3 +211,38 @@ func contactsAccept(c *gin.Context, _ *reqctx.AdminContext) {
...
 	c.JSON(http.StatusOK, contract)
 }
+
+// contactsConfirm will perform Confirm action on contacts with the given xpub ids and paymails
+// Perform confirm action on contacts godoc
+// @Summary		Confirm contacts pair
+// @Description Marks the contact entries as mutually confirmed, after ensuring the validity of the contact information for both parties.
+// @Tags		Admin
+// @Produce		json
+// @Param		[]models.AdminConfirmContactPair body []models.AdminConfirmContactPair true "Contacts data"
```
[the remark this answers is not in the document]
jakubmkowalski: `[]` removed

engine/action_contact.go
```diff
@@ -403,6 +403,81 @@ func (c *Client) upsertContact(ctx context.Context, paymailService paymailclient
...
+	}
+
+	contactB, err := getContact(ctx, paymailA, xpubB, c.DefaultModelOptions()...)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if contactA == nil || contactB == nil {
+		return nil, nil, spverrors.ErrContactsNotFound
+	}
```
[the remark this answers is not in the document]
jakubmkowalski: separated, now wraps info about user contact

dorzepowski (approved):

PR #797, opened 2024-12-05
feat(SPV-1121) [new tx flow] get tx data from OP_RETURN output

chris-4chain:
# Pull Request Checklist

### DIFF
--- a/actions/transactions/data_get.go
+++ b/actions/transactions/data_get.go
@@ -11,7 +11,7 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
-func getDataByOutpoint(c *gin.Context, _ *reqctx.UserContext) {
+func getDataByOutpoint(c *gin.Context, userContext *reqctx.UserContext) {
 	logger := reqctx.Logger(c)
 
 	var err error
@@ -23,9 +23,9 @@ func getDataByOutpoint(c *gin.Context, _ *reqctx.UserContext) {
 		return
 	}
 
-	// TODO add xpub filtering by userContext.GetXPubID()
 	data, err := reqctx.Engine(c).GetTransactionData(
 		c,
+		userContext.GetXPubID(),
 		outpoint,
 	)
 	if err != nil {
--- a/actions/transactions/data_get.go
+++ b/actions/transactions/data_get.go
@@ -31,6 +31,10 @@ func getDataByOutpoint(c *gin.Context, _ *reqctx.UserContext) {
 		spverrors.ErrorResponse(c, err, logger)
 		return
 	}
+	if data == nil {
+		spverrors.ErrorResponse(c, spverrors.ErrCouldNotFindDataOutpoint, logger)
+		return
+	}
 
 	c.JSON(http.StatusOK, response.TransactionData{
 		Data: string(data),
--- a/actions/transactions/data_get.go
+++ b/actions/transactions/data_get.go
@@ -1,13 +1,14 @@
 package transactions
 
 import (
+	"net/http"
+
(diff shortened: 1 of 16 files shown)


actions/transactions/data_get.go
```diff
@@ -0,0 +1,43 @@
...
+	var err error
+	var outpoint bsv.Outpoint
+	outpoint.TxID = c.Param("id")
+	outpoint.Vout, err = conv.StringToUint32(c.Param("vout"))
+	if err != nil {
+		spverrors.ErrorResponse(c, spverrors.ErrCannotParseParams.Wrap(err), logger)
+		return
+	}
+
+	// TODO add xpub filtering by userContext.GetXPubID()
```
chris-4chain, line 26: This is not possible right now.
We have to adjust our models (`tx-output-data`) to store such information.
- I would do that in another task.

actions/transactions/data_get_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	t.Run("try to get as admin", func(t *testing.T) {
+		// given:
+		given, then := testabilities.New(t)
+		cleanup := given.StartedSPVWallet()
+		defer cleanup()
+		client := given.HttpClient().ForAdmin()
+
+		// when:
+		res, _ := client.R().Get(fmt.Sprintf("%s/%s/0", getTransactionDataURL, fixtures.GivenTX(t).ID()))
```
mgosek-4chain, line 39: One minor question, why the second value is omitted? 🤔

mgosek-4chain (approved):

actions/transactions/data_get_test.go
chris-4chain: Hmmm I've got this pattern from @dorzepowski 🙈 
- So I don't know either 🤷

dorzepowski: Ehhh. I didn't thought it will be soooo misleading.
The client provided from testabilities is configured such a way, if there is any error (by error I mean programatic error, **not** response with status 4xx or 5xx) the test will fail. So there is no need to check that one because in case of error the line checking the error will never be called.

actions/transactions/outlines_record_test.go
```diff
@@ -57,6 +59,16 @@ func TestOutlinesRecordOpReturn(t *testing.T) {
 
 		// then:
 		then.Response(res).IsOK()
+
+		// when:
+		res, _ = client.R().
+			SetHeader("Content-Type", "application/json").
+			Get(fmt.Sprintf("%s/%s/0", getTransactionDataURL, txSpec.ID()))
```
dorzepowski, line 66: I would rather start with endpoint for listing data recorded by user.
Most important is this 'recorded by user' because, what you already mentioned, there is no possibility to do that because data is not pinned to user. But IMHO we want to index data for given user and only he can load this data as this is what he is interested in, while other users are interested in their own data.

dorzepowski (changes requested):

actions/transactions/outlines_record_test.go
chris-4chain: 1. About the 'recorded by user' thing - I've added `xpub_id` column to `Data` table. So now, `record_outline` method and getData requires `xpub_id` of a user who actually do the action.
2. I would stick to the endpoint which gets data by outpoint (txID & vout) because it seems to me that it's the most basic operation and - at the same time - allows to test the whole flow. 
3. Of course, the listings (or search) can be also introduced but - I would not return the "data" in the list items: only txIds and vouts (and maybe some other info)

PR #800, opened 2024-12-09
fix(SPV-1069) fix lint errors for "models" package

chris-4chain:
# Pull Request Checklist


models/request/opreturn/data_type.go
```diff
@@ -44,5 +46,5 @@ func (d DataType) MarshalJSON() ([]byte, error) {
 	default:
 		return nil, errors.New("invalid data type")
 	}
-	return json.Marshal(dataType)
+	return json.Marshal(dataType) //nolint:wrapcheck // MarshalJSON is run internally by json.Marshal on "DataType" object, so we don't want to wrap the error
```
ac4ch, line 49: what if u would use wrapping here ?

models/request/transaction_specification.go
```diff
@@ -45,7 +45,7 @@ func (dt *TransactionSpecification) unmarshalPartials(data []byte) (rawOutputs [
 	}
 
 	if err := json.Unmarshal(data, &temp); err != nil {
-		return nil, err
+		return nil, err //nolint:wrapcheck // unmarshalPartials is run internally by json.Unmarshal, so we don't want to wrap the error
```
ac4ch, line 48: applying same question everywhere :), why not add wrapping here `err //nolint:wrapchec` even by using fmt.Errorf or better spvwallet.err.wrap ?

chris-4chain: Because these operations run inside `json.Marshal` or `json.Unmarshal`. 

Adding error wrapping here would not provide additional value. 

These functions rely on custom implementations of the `marshaler` and `unmarshaler` interfaces to handle specific data types. When the `json.Unmarshal` function encounters an issue, it naturally returns a JSON parsing error.

Consider the following test case:
```golang
var spec *request.TransactionSpecification
err := json.Unmarshal([]byte(test.json), &spec)
require.ErrorContains(t, err, test.expectedErr)
```

In this example, the expectation is that `json.Unmarshal` will produce its own standard error messages. 
Since `request.TransactionSpecification` implements custom methods used internally by the JSON package, `UnmarshalJSON()` is invoked as part of the `json.Unmarshal` operation.

If an additional error wrapping mechanism were introduced here, it would lead to the `json.Unmarshal` call returning a custom error rather than a standard JSON parsing error. This makes the outcome less transparent and adds unnecessary complexity. 

By avoiding wrapping errors, the returned error remains clear, aligned with standard JSON operations, and easier to handle.

models/request/opreturn/data_type.go
chris-4chain: Answer below 👇

dzolt-4chain (approved):

wregulski (approved):

dorzepowski (approved):

models/request/transaction_specification.go
ac4ch: well ok ErrorIs should cover this if the Wrap is done correctly, and i think each method should wrap error not just returning it, this allow you to track errors faster, but sure i am ok with your explanation since u already got 3 approvals :)

PR #821, opened 2024-12-18
feat(SPV-1316) type42 derivation

chris-4chain:
# Pull Request Checklist

### DIFF
--- a/engine/keys/type42/derivation.go
+++ b/engine/keys/type42/derivation.go
@@ -1,0 +1,21 @@
+package type42
+
+import (
+	primitives "github.com/bitcoin-sv/go-sdk/primitives/ec"
+	"github.com/bitcoin-sv/spv-wallet/engine/spverrors"
+)
+
+var (
+	anyonePriv, _ = primitives.PrivateKeyFromBytes([]byte{1})
+)
+
+func derive(pubKey *primitives.PublicKey, derivationKey string) (*primitives.PublicKey, error) {
+	if pubKey == nil {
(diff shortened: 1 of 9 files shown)


engine/keys/type42/derivation.go
```diff
@@ -0,0 +1,21 @@
+package type42
+
+import (
+	primitives "github.com/bitcoin-sv/go-sdk/primitives/ec"
+	"github.com/bitcoin-sv/spv-wallet/engine/spverrors"
+)
+
+var (
+	anyonePriv, _ = primitives.PrivateKeyFromBytes([]byte{1})
```
dzolt-4chain, line 9: why is there anyonePriv created always from the same byte array? I thought the derivation is from the party's private key that want to create the derivation but I might be mistaken.

chris-4chain: https://bsv.brc.dev/key-derivation/0043#examples

Point 2.
>The client uses its own private key as the sender, and the public key 1 * G as the counterparty, as the counterparty is anyone.

As non-custodial wallet we don't have counterparty private key (in this case of PKI there is no counterparty at all)
- This is slightly different case - because we don't have end-to-end encryption based on shared secret but instead we want to derive public key by some `invoice number` (and we still want to use `type42` and provided functions in `go-sdk`.)
- On client side, it's possible to get `client's` private key and `anyone priv = 1` and derive `priv` with the same `invoice number` in a way that derived `priv` (on clients side) will match derived `pub` on wallet side.

dzolt-4chain (approved):

dorzepowski (approved):

PR #839, opened 2025-01-08
feat(SPV-1337): paymail capabilities test

chris-4chain:
# Pull Request Checklist


mgosek-4chain (approved):

actions/paymailserver/incoming_paymail_tx_test.go
```diff
@@ -0,0 +1,229 @@
...
+		reference     string
+		lockingScript *script.Script
+	}
+
+	// given:
+	given, then := testabilities.NewOf(givenForAllTests, t)
+	client := given.HttpClient().ForAnonymous()
+
+	// and:
+	address := fixtures.Sender.Paymails[0]
```
dorzepowski, line 32:
```suggestion
	address := fixtures.Sender.DefaultPaymail()
```

actions/paymailserver/incoming_paymail_tx_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+		// then:
+		then.Response(res).IsOK().WithJSONMatching(`{
+			"txid": "{{ .txid }}",
+			"note": "{{ .note }}"
+		}`, map[string]any{
+			"txid": txSpec.ID(),
+			"note": note,
+		})
+	})
```
dorzepowski, line 117: Maybe it would be good to check the balance here also?

actions/paymailserver/stateless_capabilities_test.go
```diff
@@ -0,0 +1,194 @@
...
+		}`)
+	})
+
+	t.Run("Public profile", func(t *testing.T) {
+		// given:
+		given, then := testabilities.NewOf(givenForAllTests, t)
+		client := given.HttpClient().ForAnonymous()
+
+		// and:
+		address := fixtures.Sender.Paymails[0]
```
dorzepowski, line 52:
```suggestion
		address := fixtures.Sender.DefaultPaymail()
```

engine/tester/jsonrequire/funcs.go
```diff
@@ -24,6 +27,18 @@ func matchID64() string {
 	return `/^[a-zA-Z0-9]{64}$/`
 }
 
+func matchHexWithLength(length int) string {
+	return fmt.Sprintf(`/^[a-f0-9]{%d}$/`, length)
```
dorzepowski, line 31: A-F can be also in hex (if it is upper case)

server/server.go
```diff
@@ -107,8 +108,7 @@ func (s *Server) Handlers() *gin.Engine {
 func setupServerRoutes(appConfig *config.AppConfig, spvWalletEngine engine.ClientInterface, ginEngine *gin.Engine) {
 	handlersManager := handlers.NewManager(ginEngine, appConfig)
 	actions.Register(handlersManager)
-
-	spvWalletEngine.GetPaymailConfig().RegisterRoutes(ginEngine)
+	paymailserver.Register(spvWalletEngine.GetPaymailConfig().Configuration, ginEngine)
```
dorzepowski, line 111: Why not:
```suggestion
	paymailserver.Register(spvWalletEngine.GetPaymailConfig(), ginEngine)
```

engine/tester/jsonrequire/getter.go
```diff
@@ -0,0 +1,36 @@
...
+	return &Getter{t: t, data: data}
+}
+
+// GetString returns a string value from the data.
+func (g *Getter) GetString(xpath string) string {
+	value := getByXPath(g.t, g.data, xpath)
+
+	strValue, ok := value.(string)
+	if !ok {
+		require.Fail(g.t, "Value is not a string")
```
dorzepowski, line 32: maybe it would be good to add some more context, like
```suggestion
		require.Fail(g.t, "Value on xpath %s is not a string, it is %T (%v)", xpath, value, value)
```

dorzepowski (changes requested):

actions/paymailserver/incoming_paymail_tx_test.go
chris-4chain: But the balance check will change in new approach so I don't think it's beneficial to add it now (in the old way)

engine/tester/jsonrequire/funcs.go
chris-4chain: It's nice you also reviewing my regexes 🆒 😄

server/server.go
chris-4chain: That ☝️ was my first attempt but didn't work that way - because of the object wrapping.

engine/tester/jsonrequire/getter.go
chris-4chain: Thanks :) I used your suggestion 👍

engine/tester/jsonrequire/funcs.go
chris-4chain: Updated 👍

actions/paymailserver/incoming_paymail_tx_test.go
chris-4chain: Thanks for `DefaultPaymail` suggestion - I didn't know about that 🙈 

But I did more changes about the paymails addresses. 
- I've switched from `Sender` to `RecipientInternal` to reflect who is who:
```golang
senderPaymail := "bob@external.com"
recipientPaymail := fixtures.RecipientInternal.DefaultPaymail()
```

FYI: I've applied the same change for RawTx, Beef and for "stateless capabilities"

actions/paymailserver/stateless_capabilities_test.go
chris-4chain: I've changed to `fixtures.RecipientInternal.DefaultPaymail()` in all occurrences.

engine/tester/jsonrequire/getter.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Getter allows to get values from the JSON data.
+type Getter struct {
+	t    testing.TB
+	data map[string]any
+}
+
+// NewGetterWithJSON creates a new Getter based on JSON string
+func NewGetterWithJSON(t testing.TB, jsonString string) *Getter {
+	var data map[string]any
```
wregulski, line 19: I like it 👍

actions/paymailserver/incoming_paymail_tx_test.go
chris-4chain: Good idea - done 😄

engine/tester/jsonrequire/funcs.go
```diff
@@ -24,6 +27,18 @@ func matchID64() string {
 	return `/^[a-zA-Z0-9]{64}$/`
 }
 
+func matchHexWithLength(length int) string {
+	return fmt.Sprintf(`/^[a-fA-F0-9]{%d}$/`, length)
+}
+
+func matchHex() string {
+	return `/^[a-f0-9]+$/`
```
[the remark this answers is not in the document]
chris-4chain: Done 👍

engine/tester/jsonrequire/funcs.go
```diff
@@ -24,6 +27,18 @@ func matchID64() string {
...
+}
+
+func matchAddress() string {
+	return `/^(1|m)[a-km-zA-HJ-NP-Z1-9]{33}$/`
```
wregulski, line 39: I spent a while on that, don't kill me 🙈 
I guess this is the correct one for addresses if we also want to cover testnet cases.

Source:
[Bitcoin-SV Wiki](https://wiki.bitcoinsv.io/index.php/Technical_background_of_Bitcoin_addresses?utm_source=chatgpt.com)

```go
func matchAddress() string {
    return `^(1[a-km-zA-HJ-NP-Z1-9]{24,33}|[mn][a-km-zA-HJ-NP-Z1-9]{25,33})$`
}
```

engine/tester/jsonrequire/funcs.go
wregulski: Shouldn't we include uppercase hex "letters"?
```go
func matchHex() string {
    return `^[a-fA-F0-9]+$`
}
```

engine/tester/jsonrequire/funcs.go
```diff
@@ -24,6 +27,18 @@ func matchID64() string {
 	return `/^[a-zA-Z0-9]{64}$/`
 }
 
+func matchHexWithLength(length int) string {
```
wregulski, line 30: I think that those slashes **/** are redundant, we can use:
```go
func matchHexWithLength(length int) string {
    return fmt.Sprintf(`^[a-fA-F0-9]{%d}$`, length)
}
```

chris-4chain: Slashes are not redundant - because they have meaning for `jsonrequire.assertJSONWithPlaceholders` 
- this is an indicator to treat the string as regex in my placeholder matching engine.

engine/tester/jsonrequire/funcs.go
chris-4chain: Done :) 
@dorzepowski was faster 😄

wregulski: 🥈

engine/tester/jsonrequire/funcs.go
wregulski: Okay, understood 👍

actions/paymailserver/incoming_paymail_tx_test.go
chris-4chain: Ok, ok 😄 
Done :)

engine/tester/jsonrequire/funcs.go
wregulski: Then, regarding to the specification I would rather go with this regexp:
```go
func matchAddress() string {
    return `/^1[a-km-zA-HJ-NP-Z1-9]{24,33}$/`
}
```

because currently (1|m) "m" symbol points to the testnet.

wregulski (approved):

engine/tester/jsonrequire/funcs.go
chris-4chain: Refresh the github and check that your proposition is exactly the same as mine 🙈

dorzepowski (approved):

PR #842, opened 2025-01-08
feat(SPV-1295): testabilities for contacts admin

dzolt-4chain:
# Pull Request Checklist


engine/contact/contact_test.go
```diff
@@ -0,0 +1,150 @@
...
+	tests := map[string]struct {
+		contactPaymail   string
+		creatorPaymail   string
+		fullName         string
+		metadata         *engine.Metadata
+		expectedError    error
+		expectedStatus   engine.ContactStatus
+		expectedFullName string
+	}{
+		"Happy path without metadata": {
```
dorzepowski, line 24:
```suggestion
		"Create contact without metadata": {
```

engine/contact/contact_test.go
```diff
@@ -0,0 +1,150 @@
...
+			contactPaymail:   fixtures.RecipientExternal.DefaultPaymail(),
+			creatorPaymail:   fixtures.Sender.DefaultPaymail(),
+			fullName:         "John Doe",
+			metadata:         nil,
+			expectedError:    nil,
+			expectedStatus:   engine.ContactNotConfirmed,
+			expectedFullName: "John Doe",
+		},
+		"Happy path with metadata": {
```
dorzepowski, line 33:
```suggestion
		"Create contact with metadata": {
```

engine/contact/contact_test.go
```diff
@@ -0,0 +1,150 @@
...
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			//given:
+			given, then := testabilities.New(t)
+
+			service, cleanup := given.Engine()
+			defer cleanup()
+
+			//when:
+			res, err := service.AdminCreateContact(context.Background(), tt.contactPaymail, tt.creatorPaymail, tt.fullName, tt.metadata)
```
dorzepowski, line 56: I feel that is unit tests instead of `res` (and later `WithResponse`) we should use word `contact` or something because this isn't the endpoint test and is rather focused on a specific feature, then the API with some responses.

engine/contact/contact_test.go
```diff
@@ -0,0 +1,150 @@
...
+		defer cleanup()
+
+		//and:
+		given.PaymailClient().WillRespondOnCapability(paymail.BRFCPki).WithInternalServerError()
+
+		//when:
+		res, err := service.AdminCreateContact(context.Background(), fixtures.RecipientExternal.DefaultPaymail(), fixtures.Sender.DefaultPaymail(), "John Doe", nil)
+
+		//then:
+		then.ErrorIs(err, spverrors.ErrGettingPKIFailed).WithNilResponse(res)
```
dorzepowski, line 79: If you would go with structure more like this:
```suggestion
		then.ContactCreated(res).WithError(err).ThatIs(spverrors.ErrGettingPKIFailed)
```
then in `WithError` you could require `res` (contact) to be `nil`, as this is what we always expect when there is an error.

ℹ️ You can find example of something like that in [create_draft_test.go](https://github.com/bitcoin-sv/spv-wallet/blob/3312ad9f92c0acb11723db02e18850535b08b549/engine/transaction/outlines/create_draft_test.go#L51)

engine/contact/contact_test.go
```diff
@@ -0,0 +1,150 @@
...
+			given, then := testabilities.New(t)
+
+			service, cleanup := given.Engine()
+			defer cleanup()
+
+			//when:
+			res, err := service.AdminCreateContact(context.Background(), tt.contactPaymail, tt.creatorPaymail, tt.fullName, tt.metadata)
+
+			//then:
+			then.NoError(err).WithResponse(res).WithStatus(tt.expectedStatus).WithFullName(tt.expectedFullName)
```
dorzepowski, line 59: IMHO you should also 
1. check the PubKey - if it matches the one from external paymail server. 
2. And maybe also check the Paymail field that it is paymail of contact
3. and maybe could add an assertion that contact was created for user, and pass the fixtures.Sender and check if OwnerXpubID is matching user.XPubID

to sum up, I would see it like this:
```suggestion
			then.ContactCreated(contact).WithoutError(err)
			  .ForUser(fixtures.Sender) // Maybe it could be named InContactBookOf(fixtures.Sender)
			  .WithContactPaymailOf(fixtures.RecipientExternal)
			  .WithPubKey(fixtures.RecipientExternalPKI)
			  .WithStatus(tt.expectedStatus)
			  .WithFullName(tt.expectedFullName)
```

alternatively it could be something like this:
```suggestion
			then.ContactCreated(contact).WithoutError(err)
			  .ForUser(fixtures.Sender)  // Maybe it could be named InContactBookOf(fixtures.Sender)
			  .ToCounterparty(fixtures.RecipientExternal) // and here we check the paymail and PubKey
			  .WithStatus(tt.expectedStatus)
			  .WithFullName(tt.expectedFullName)
```

engine/contact/contact_test.go
```diff
@@ -0,0 +1,150 @@
...
+func Test_ClientService_AdminCreateContact_Fail(t *testing.T) {
+	tests := map[string]struct {
+		contactPaymail   string
+		creatorPaymail   string
+		fullName         string
+		expectedError    error
+		expectedStatus   engine.ContactStatus
+		expectedFullName string
+	}{
+		"Edge case: Creator paymail not found": {
```
dorzepowski, line 92: I would rather go with:
- `should fail when ...`
- `should return error when ...`
- `return error when ...`

instead of `Edge case: ` 

It's actually not even an edge case, it's rather error case 🙂

engine/contact/contact_test.go
```diff
@@ -0,0 +1,150 @@
...
+			expectedFullName: "",
+		},
+		"Edge case: missing creator paymail": {
+			contactPaymail: fixtures.RecipientExternal.DefaultPaymail(),
+			creatorPaymail: "",
+			fullName:       "John Doe",
+			expectedError:  spverrors.ErrMissingContactCreatorPaymail,
+		},
+		"Edge case: missing contact full name": {
+			contactPaymail: fixtures.RecipientExternal.DefaultPaymail(),
```
dorzepowski, line 107: there is no test for missing `contactPaymail`

engine/contact/testabilities/assert_contact.go
```diff
@@ -0,0 +1,69 @@
...
+	return a
+}
+
+func (a *assert) WithNilResponse(response *engine.Contact) ContactFailureAssertion {
+	a.require.Nil(response, "unexpected response")
+	return a
+}
+
+func (a *assert) WithStatus(status engine.ContactStatus) ContactSuccessAssertion {
+	a.require.Equal(status, a.response.Status, fmt.Sprintf("expected status: %s, actual: %s", status, a.response.Status))
```
dorzepowski, line 62: There is no need for `fmt.Sprintf` because the methods in `require` nad `assert` are accepting `string` which is string template as the and the args to this template as the parameters

In other words, this should work:
```suggestion
	a.require.Equal(status, a.response.Status, "expected status: %s, actual: %s", status, a.response.Status)
```

Additionally maybe it's enough to just put string `"Contact has invalid status"`, because `require.Equal` will show the difference between expected and received values.

👇 same thing for the WithFullName

engine/contact/testabilities/assert_contact.go
```diff
@@ -0,0 +1,69 @@
...
+	return a
+}
+
+func (a *assert) WithFullName(fullName string) ContactSuccessAssertion {
+	a.require.Equal(fullName, a.response.FullName, fmt.Sprintf("expected fullName: %s, actual: %s", fullName, a.response.FullName))
```
dorzepowski, line 67: In checks of the specific fields in contact (like WithFullName, WithStatus, etc.) I would prefer to use package `assert` instead of `require`.

engine/contact/testabilities/assert_contact.go
```diff
@@ -0,0 +1,69 @@
...
+	a.response = response
+	return a
+}
+
+func (a *assert) WithNilResponse(response *engine.Contact) ContactFailureAssertion {
+	a.require.Nil(response, "unexpected response")
+	return a
+}
+
+func (a *assert) WithStatus(status engine.ContactStatus) ContactSuccessAssertion {
```
dorzepowski, line 61: because in those tests we're always expect to have status `ContactNotConfirmed` maybe it would be good to have (and call) method `NotConfirmed()` or `AsNotConfirmed()`.
so something like that:
```golang
func (a *assert) AsNotConfirmed() {
   return a.WithStatus(engine.ContactNotConfirmed)
}
```

engine/contact/contact_test.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+func Test_ClientService_AdminCreateContact_Success(t *testing.T) {
+	tests := map[string]struct {
+		contactPaymail   string
+		creatorPaymail   string
+		fullName         string
+		metadata         *engine.Metadata
+		expectedError    error
+		expectedStatus   engine.ContactStatus
```
dorzepowski, line 21: When created, we always expect the status to be `ContactNotConfirmed` so I believe that this field is not needed.

engine/contact/contact_test.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func Test_ClientService_AdminCreateContact_Fail(t *testing.T) {
+	tests := map[string]struct {
+		contactPaymail   string
+		creatorPaymail   string
+		fullName         string
+		expectedError    error
+		expectedStatus   engine.ContactStatus
+		expectedFullName string
```
dorzepowski, line 90: In error cases you don't need the expected status and full name as actually the contact is nil so it doesn't has those fields 🙂

engine/contact/contact_test.go
dorzepowski, line 1: It would be good to have also endpoint tests (but it could be done in a separate PR)

engine/contact/contact_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"testing"
+
+	"github.com/bitcoin-sv/go-paymail"
+	"github.com/bitcoin-sv/spv-wallet/engine"
+	"github.com/bitcoin-sv/spv-wallet/engine/contact/testabilities"
+	"github.com/bitcoin-sv/spv-wallet/engine/spverrors"
+	"github.com/bitcoin-sv/spv-wallet/engine/tester/fixtures"
+)
+
+func Test_ClientService_AdminCreateContact_Success(t *testing.T) {
```
dorzepowski, line 14: You should remove the equivalent tests from the engine/contact_service_test.go

dorzepowski (changes requested):
Great job, 
For me, it reads much better then the previous approach.

engine/contact/testabilities/assert_contact.go
```diff
@@ -0,0 +1,69 @@
...
+func (a *assert) NoError(err error) ContactSuccessAssertion {
+	a.require.NoError(err, "Record transaction outline has error")
+	return a
+}
+
+func (a *assert) ErrorIs(err, expectedError error) ContactFailureAssertion {
+	require.Error(a.t, err, "Record transaction outline has no error")
+	require.ErrorIs(a.t, err, expectedError, "Record transaction outline has wrong error")
+	return a
+}
```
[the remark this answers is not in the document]
dzolt-4chain: oh you're right :D

engine/contact/contact_test.go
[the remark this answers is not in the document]
dzolt-4chain: sure :D

engine/contact/contact_test.go
```diff
@@ -0,0 +1,150 @@
...
+		expectedFullName string
+	}{
+		"Happy path without metadata": {
+			contactPaymail:   fixtures.RecipientExternal.DefaultPaymail(),
+			creatorPaymail:   fixtures.Sender.DefaultPaymail(),
+			fullName:         "John Doe",
+			metadata:         nil,
+			expectedError:    nil,
+			expectedStatus:   engine.ContactNotConfirmed,
+			expectedFullName: "John Doe",
```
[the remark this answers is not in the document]
dzolt-4chain: you're right. changed

engine/contact/contact_test.go
dzolt-4chain: makes sense changed :D

engine/contact/contact_test.go
dzolt-4chain: I didn't change original names :D Will change them

engine/contact/testabilities/assert_contact.go
dzolt-4chain: I would leave it as is, because I believe in the end we will want to move to testabilities in the whole code base and there are other methods that test different statuses like [here](https://github.com/bitcoin-sv/spv-wallet/blob/main/engine/contact_service_test.go#L134) and [here](https://github.com/bitcoin-sv/spv-wallet/blob/main/engine/contact_service_test.go#L213)

engine/contact/contact_test.go
dzolt-4chain: removed

engine/contact/contact_test.go
dzolt-4chain: added

engine/contact/contact_test.go
dzolt-4chain: done 🚀

engine/contact/testabilities/assert_contact.go
dzolt-4chain: changed to assert

engine/contact/contact_test.go
dzolt-4chain: done :D

chris-4chain (approved):

engine/contact/contact_test.go
```diff
@@ -0,0 +1,177 @@
...
+			contact, err := service.AdminCreateContact(context.Background(),
+				tt.contactPaymail,
+				tt.creatorPaymail,
+				tt.fullName,
+				tt.metadata,
+			)
+
+			//then:
+			then.
+				Contact(contact).
```
dorzepowski, line 56: ### minor
```suggestion
				Created(contact).
```

dorzepowski (changes requested):

engine/contact/contact_test.go
dzolt-4chain: To me it seams clearer when its Contact but to be consistent with previous implementations I'm changing it here

engine/contact/testabilities/assert_contact.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (a *assertion) WithNoError(err error) ContactSuccessAssertion {
+	a.require.NotNil(a.contact, "unexpected nil response")
+	a.require.NoError(err, "record transaction outline has error")
+	return a
+}
+
+func (a *assertion) ThatIs(expectedError error) ContactFailureAssertion {
+	a.require.Nil(a.contact, "unexpected response")
```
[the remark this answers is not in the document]
dzolt-4chain: yeah, also just noticed it thanks for pointing it out :D

dorzepowski (approved):

PR #850, opened 2025-01-13
bug(SPV-1342) admin_endpoint_for_paginated_response_doesnt_map_status_and_direction_in_response

ac4ch:
# Pull Request Checklist


mgosek-4chain (approved):

mappings/transaction.go
```diff
@@ -38,29 +38,7 @@ func MapToTransactionContract(t *engine.Transaction) *response.Transaction {
...
-		NumberOfOutputs: t.NumberOfOutputs,
-		DraftID:         t.DraftID,
-		TotalValue:      t.TotalValue,
-		Outputs:         t.XpubOutputValue,
-	}
-
-	processMetadata(t, t.XPubID, &model)
-
-	return &model
+	return MapToTransactionContract(t)
```
dzolt, line 41: LGTM, however if this is doing exactly the same as `MapToTransactionContract` maybe we don't need this method `MapToTransactionContractForAdmin` and we should use `MapToTransactionContract` everywhere?

dzolt: Now that I'm looking at it I think we are now missing outputs for admin model, so actually I would leave the method and after `MapToTransactionContract` we need to also map outputs there for the output.

dzolt-4chain (approved):

PR #866, opened 2025-01-22
refactor(SPV-1404): adjust tx outline to type42 changes.

dorzepowski:
# Pull Request Checklist


engine/transaction/outlines/testabilities/assert_outline_output.go
```diff
@@ -0,0 +1,79 @@
...
+		a.assert.Equal(receiver, a.annotation.Paymail.Receiver, "Output %d has invalid paymail receiver", a.index)
+	}
+	return a
+}
+
+func (a *txOutputAssertion) HasSender(sender string) TransactionOutlinePaymailOutputAssertion {
+	a.t.Helper()
+	if a.annotation.Paymail != nil {
+		a.assert.Equal(sender, a.annotation.Paymail.Sender, "Output %d has invalid paymail sender", a.index)
+	}
```
chris-4chain, line 69: Same situation as ☝️

engine/transaction/outlines/testabilities/assert_outline_output.go
```diff
@@ -0,0 +1,79 @@
...
+	return a
+}
+
+func (a *txOutputAssertion) HasReference(reference string) TransactionOutlinePaymailOutputAssertion {
+	a.t.Helper()
+	if a.annotation.Paymail != nil {
+		a.assert.Equal(reference, a.annotation.Paymail.Reference, "Output %d has invalid paymail reference", a.index)
+	}
```
chris-4chain, line 77: ☝️

engine/transaction/outlines/testabilities/assert_outline_transaction.go
```diff
@@ -63,25 +68,29 @@ type assertion struct {
...
+	a.t.Helper()
 	a.assert.ErrorIs(a.err, expectedError)
 }
 
 // WithNoError checks if there was no error and result is not nil. It also checks if BEEF hex is parseable.
 func (a *assertion) WithNoError(err error) SuccessfullyCreatedTransactionOutlineAssertion {
+	a.t.Helper()
 	a.require.NoError(err, "Creation of transaction outline has finished with error")
-	a.require.NotNil(a.txOutline, "Transaction outline should be created if there is no error")
+	a.require.NotNil(a.txOutline, "The result is nil although there was no error")
```
chris-4chain, line 93:
```suggestion
	a.require.NotNil(a.txOutline, "The result is nil even though there was no error")
```

engine/transaction/outlines/testabilities/mock_paymail_address_service.go
```diff
@@ -0,0 +1,42 @@
...
+	return false, nil
+}
+
+func (m *mockPaymailAddressService) GetDefaultPaymailAddress(_ context.Context, userID string) (string, error) {
+	for _, user := range m.users {
+		if user.ID() == userID {
+			if user.DefaultPaymail() != "" {
+				return user.DefaultPaymail(), nil
+			}
+		}
```
chris-4chain, line 39:
```suggestion
		if user.ID() == userID && user.DefaultPaymail() != "" {
			return user.DefaultPaymail(), nil
		}
```

engine/paymailaddress/paymail_address_service.go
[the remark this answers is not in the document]
dorzepowski: IMHO the name of the file doesn't need to follow package name,
packages have specific requirements, for example it shouldn't have `_`, while file name can have it and therefore can have more readable name.

engine/database/repository/paymails.go
```diff
@@ -34,3 +34,20 @@ func (p *Paymails) Get(ctx context.Context, alias, domain string) (*database.Pay
 
 	return &paymail, nil
 }
+
+// GetDefault returns a default paymail for user.
```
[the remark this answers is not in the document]
mgosek-4chain: We could return a sentinel error to clearly indicate that the resource was not found then handle such error as 404 response. 🤔

engine/paymailaddress/paymail_address_service.go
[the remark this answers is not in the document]
mgosek-4chain: I agree with @dorzepowski. I think it might be a good idea to revisit and update our naming convention list.

engine/transaction/annotation.go
```diff
@@ -7,9 +7,20 @@ import (
...
+	Inputs  InputAnnotations
 	Outputs OutputsAnnotations
 }
 
+// InputAnnotations represents the metadata for chosen inputs. The key is the index of the input.
+type InputAnnotations map[int]*InputAnnotation
+
+// InputAnnotation represents the metadata for the input.
+type InputAnnotation struct {
+}
```
mgosek-4chain, line 19: Could you explain why this structure is empty? 🤔

engine/transaction/outlines/evaluation_context.go
```diff
@@ -8,27 +8,41 @@ import (
...
 
+// evaluationContext is a context for the evaluation of a transaction outline specification.
+type evaluationContext interface {
+	context.Context
+	XPubID() string
+	UserID() string
+	Log() *zerolog.Logger
+	Paymail() paymail.ServiceClient
+	PaymailAddressService() paymailaddress.Service
+}
```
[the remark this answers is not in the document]
mgosek-4chain: Good question 🤔 . Based on the general approach `context` should typically be passed as first parameter of each method.

engine/transaction/outlines/evaluation_context.go
```diff
@@ -8,27 +8,41 @@ import (
...
+
 type ctx struct {
```
mgosek-4chain, line 21: I would recommend avoid naming custom types as `ctx` or `context`, as it can be misleading and might cause confusion with `ctx` from the `context` package.

engine/transaction/outlines/output_op_return.go
```diff
@@ -9,14 +9,13 @@ import (
...
 	txerrors "github.com/bitcoin-sv/spv-wallet/engine/transaction/errors"
-	"github.com/bitcoin-sv/spv-wallet/engine/transaction/outlines/internal/evaluation"
 	"github.com/bitcoin-sv/spv-wallet/models/request/opreturn"
 )
 
 // OpReturn represents an OP_RETURN output specification.
 type OpReturn opreturn.Output
 
-func (o *OpReturn) evaluate(evaluation.Context) (annotatedOutputs, error) {
+func (o *OpReturn) evaluate(evaluationContext) (annotatedOutputs, error) {
```
mgosek-4chain, line 18: It might be better to create the child context from the root context passed as `ctx` parameter. However, I noticed that `evaluationContext` doesn't seem to be used anywhere in the `evaluate` method definition. 🤔

engine/transaction/outlines/output_paymail.go
```diff
@@ -10,15 +10,14 @@ import (
...
-	"github.com/bitcoin-sv/spv-wallet/engine/transaction/outlines/internal/evaluation"
 	paymailreq "github.com/bitcoin-sv/spv-wallet/models/request/paymail"
 	"github.com/bitcoin-sv/spv-wallet/models/transaction/bucket"
 )
 
 // Paymail represents a paymail output
 type Paymail paymailreq.Output
 
-func (p *Paymail) evaluate(ctx evaluation.Context) (annotatedOutputs, error) {
+func (p *Paymail) evaluate(ctx evaluationContext) (annotatedOutputs, error) {
```
mgosek-4chain, line 20: Like in the previous comment about using child context 😸

engine/transaction/outlines/testabilities/assert_outline_output.go
```diff
@@ -0,0 +1,79 @@
...
+	a.require.NotNil(a.annotation, "Output %d has no annotation", a.index)
+	a.assert.NotNil(a.annotation.Paymail, "Output %d is not a paymail output", a.index)
+	return a
+}
+
+func (a *txOutputAssertion) HasReceiver(receiver string) TransactionOutlinePaymailOutputAssertion {
+	a.t.Helper()
+	if a.annotation.Paymail != nil {
+		a.assert.Equal(receiver, a.annotation.Paymail.Receiver, "Output %d has invalid paymail receiver", a.index)
+	}
```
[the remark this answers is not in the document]
mgosek-4chain: Can we simplify this into the following assertion?

```
	a.assert.NotNil(a.annotation.Paymail)
	a.assert.Equal(receiver, a.annotation.Paymail.Receiver, "Output %d has invalid paymail receiver", a.index)
```

engine/transaction/outlines/testabilities/mock_paymail_address_service.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+}
+
+func (m *mockPaymailAddressService) HasPaymailAddress(_ context.Context, userID string, address string) (bool, error) {
+	for _, user := range m.users {
+		if user.ID() == userID {
+			return slices.Contains(user.Paymails, address), nil
+		}
+	}
+	return false, nil
```
mgosek-4chain, line 30: Consider returning just a boolean value instead of a tuple (bool, error), since the error is always returned as nil in both return statements. 🤔

engine/database/repository/paymails.go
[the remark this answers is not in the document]
dorzepowski: I'm returning now errors as is, and handling it in the service

engine/paymailaddress/paymail_address_service.go
```diff
@@ -3,51 +3,44 @@ package paymailaddress
...
+func NewService(repo *repository.Paymails) Service {
+	return &service{repo: repo}
 }
 
-// HasPaymailAddress checks if the given address belongs to a given xPubId.
-func (s *service) HasPaymailAddress(ctx context.Context, xPubID string, address string) (bool, error) {
-	paymailXpubID, err := s.getXPubIDByPaymailAddress(ctx, address)
+// HasPaymailAddress checks if the given address belongs to a given User.
+func (s *service) HasPaymailAddress(ctx context.Context, userID string, address string) (bool, error) {
+	alias, domain, _ := paymail.SanitizePaymail(address)
```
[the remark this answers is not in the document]
dorzepowski: for now, I will check it as you wrote, 
But in the future we should think about some solution for paymail to have single easy way of validating it and passing as already validated.

engine/paymailaddress/paymail_address_service.go
```diff
@@ -3,51 +3,44 @@ package paymailaddress
...
 	if err != nil {
 		return false, err
 	}
-	return paymailXpubID == xPubID, nil
+
+	if pm == nil {
+		return false, nil
+	}
+
+	return pm.UserID == userID, nil
```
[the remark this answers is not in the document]
dorzepowski: I changed it to find for user

engine/transaction/errors/errors.go
```diff
@@ -24,6 +24,9 @@ var (
 	// ErrTxOutlineSenderPaymailAddressNoDefault is when it is not possible to determine the default address for the sender.
 	ErrTxOutlineSenderPaymailAddressNoDefault = models.SPVError{Code: "error-tx-spec-paymail-address-no-default", Message: "cannot choose paymail address of the sender", StatusCode: 400}
 
+	// ErrTxOutlineInsufficientFunds is returned when user has not enough BSV in UTXOs to fulfil the transaction.
```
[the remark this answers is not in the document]
dorzepowski: it is fight between UK and US, I will go with more peaceful word 😄

engine/transaction/outlines/testabilities/assert_outline_transaction.go
dorzepowski: nope! 
backaway

actions/transactions/internal/mapping/outline/outline_request_to_engine.go
```diff
@@ -17,9 +16,9 @@ import (
 type Request request.TransactionSpecification
 
 // ToEngine converts a transaction outline request model to the engine model.
-func (tx Request) ToEngine(xPubID string) (*outlines.TransactionSpec, error) {
+func (tx Request) ToEngine(userID string) (*outlines.TransactionSpec, error) {
 	spec := &outlines.TransactionSpec{
-		XPubID: xPubID,
+		UserID: userID,
```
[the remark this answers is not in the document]
dorzepowski: It's used to [put into evaluationContext](https://github.com/bitcoin-sv/spv-wallet/blob/bfde677e143d36b04b3dda363bcb3678aa12fa17/engine/transaction/outlines/transaction_outlines_service.go#L48) and then it is used for example by [output_paymail](https://github.com/bitcoin-sv/spv-wallet/blob/bfde677e143d36b04b3dda363bcb3678aa12fa17/engine/transaction/outlines/output_paymail.go#L94)

engine/transaction/annotation.go
dorzepowski: The case is that:
I started with trying to fill the outline with UTXO,
I prepared some code (actually almost done), 
then I figured out that I need to use UserID,
then I figured out that PaymailAddress Service needs also to be changed because it needs to use new tables.
So I decide to make a step back, do a refactoring and adjust the code to those new tables (like user and paymail)
And to do that, I removed the things regarding Inputs, but it seems that I forget to remove this particular strucure.

If you don't mind, I would keep it, to make it simpler to restore the work regarding inputs.

engine/transaction/outlines/evaluation_context.go
dorzepowski: Can you then suggest some name?

engine/transaction/outlines/testabilities/mock_paymail_address_service.go
dorzepowski: This is a mock, which is used to implement interface in **tests**, so I cannot remove error, because it won't match the interface.

engine/transaction/outlines/testabilities/assert_outline_output.go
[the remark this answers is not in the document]
dorzepowski: **tl; tr;**
I see that it was confusing, I made a change now it's simpler.


And the context of that code and the change:
- because we use interfaces, there wasn't a way to call `HasReceiver`, `HasSender`, `HasReference`, before making a call to `IsPaymail` method.
- `IsPaymail` was checking if the `a.annotationsPaymail` is not nil
- BUT it checks it with assert, instead of require (because I thought it could be called for example before assertions about other parts of tx)
- because assert will already fail, 
   - I didn't want to have the rest of assertion panic with nil
   - And I didn't want to print 4 messages that `a.annotation.Paymail is nil`
- It seems that it was confusing, so
   - I changed the assertion on `a.annotationsPaymail` in `IsPaymail` to use require
   - so now, if it's nil, the test will fail instantly, so the rest of assertions regarding Paymail output won't run
   - so we don't need to check if `a.annotation.Paymail` is `nil` in those Paymail output assertions, because for sure it isn't if the runtime can call those methods.

engine/transaction/outlines/testabilities/assert_outline_output.go
dorzepowski: ☝️ answered

engine/transaction/outlines/testabilities/assert_outline_output.go
dorzepowski: ☝️ answered

engine/transaction/outlines/output_op_return.go
dorzepowski: It is forced by interface, here that's why it's not used.

engine/transaction/outlines/evaluation_context.go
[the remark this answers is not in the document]
dorzepowski: To not split the discussion into several comments, I will try to address them all here.
So the issues that you're pointing out are:

1. I'm using interface instead of just `ctx` struct although both are private and in the same package

It's not necessary, but I somehow want to enforce that we remember to fill all the required dependencies, that can be used by any of the outputs or inputs.
Unfortunately I don't have better idea for making a dependency injection to polymorphic components of the spec, that are building transaction.
But I will change it to use plain struct, because you're right that making this interface is pointless

2. name of the struct `ctx` which can be misleading

I choose that name to make a problem for a person who would try to create the struct without the function newTransactionContext. And I made this to ensure that all the deps are filled and nothing was forgotten. (this function already helped me when I was switching from userID to xpubID)

3. It might be better to create the child context from the root context passed as ctx parameter. 

The idea of evaluation context is to pass all the needed things to components that are building outputs and inputs.
Maybe it isn't the best idea, maybe there is better way to do that, but I'm not convinced to the solution of assigning the services in context.Context as this seems to make thing harder to be guarded by compiler.

engine/transaction/outlines/evaluation_context.go
dorzepowski: I'm moving discussion to: https://github.com/bitcoin-sv/spv-wallet/pull/866#discussion_r1928301414

engine/transaction/outlines/output_paymail.go
dorzepowski: I'm moving discussion to: https://github.com/bitcoin-sv/spv-wallet/pull/866#discussion_r1928301414

engine/database/repository/paymails.go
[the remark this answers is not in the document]
chris-4chain: If so, we need to adjust other repositories' methods to this approach.

engine/transaction/errors/errors.go
```diff
@@ -24,6 +24,9 @@ var (
 	// ErrTxOutlineSenderPaymailAddressNoDefault is when it is not possible to determine the default address for the sender.
 	ErrTxOutlineSenderPaymailAddressNoDefault = models.SPVError{Code: "error-tx-spec-paymail-address-no-default", Message: "cannot choose paymail address of the sender", StatusCode: 400}
 
+	// ErrTxOutlineInsufficientFunds is returned when user has not enough BSV in UTXOs to fund the transaction.
+	ErrTxOutlineInsufficientFunds = models.SPVError{Code: "tx-outline-not-enough-funds", Message: "not enough funds to make the transaction", StatusCode: 422}
```
chris-4chain, line 28: It seems to be unused

chris-4chain (approved):

mgosek-4chain (approved):

PR #876, opened 2025-01-28
feat: create example for openapi3

pawellewandowski98:
# Description

- This is an example of how the API interface can be build in SPV-Wallet

Run `go generate ./...` or Make api. What will happen?

1. One, merged yaml file will be generated with whole API specification
  - It will merge `api_base.yaml` `api_user.yaml` `api_admin.yaml` and `components_models.yaml` `components_errors.yaml`
2. API specification will be generated using `oapi-codegen`

# Pull Request Checklist


wregulski:
nice 👍

Makefile
```diff
@@ -50,3 +50,9 @@ test-all-db-ci: ## Runs all tests including embedded database tests (CI)
 	cd models/ && go test ./... -coverprofile=coverage.txt -covermode=atomic -tags="$(GO_BUILD_TAGS) database_tests"
 	tail -n +2 models/coverage.txt >> coverage.txt
 	tail -n +2 engine/coverage.txt >> coverage.txt
+
+.PHONY: api
+api: ## Generate API
+	@echo "generating API..."
+	@go generate ./...
+	@echo "API generation completed, go to /api/api.go to see the generated API"
```
chris-4chain, line 58: Hmm, consider `tools` pattern - as they recommend in documentation:
https://github.com/oapi-codegen/oapi-codegen?tab=readme-ov-file#install

I also used it here: 
https://github.com/bitcoin-sv/spv-wallet/pull/859/files#diff-1afd39135576606ef3cad30157a9dd3fc749b2be429e8d1778ad95d2c93c75a1

api/components/models.yaml
```diff
@@ -0,0 +1,66 @@
...
+      properties:
+        id:
+          type: integer
+          format: uint64
+          nullable: false
+          description: Example of admin request body
+          example: 1
+      additionalProperties: false
+
+    #######################################################  MODELS
```
chris-4chain, line 24: We should consider if we'd like to store models in separated package.

api/configs/api.yaml
```diff
@@ -0,0 +1,8 @@
+package: api
+generate:
+  gin-server: true
+  models: true
+output: gen.api.go
+#import-mapping:
+#  errors.yaml: "-"
+#  /models/models.yaml: "-"
```
chris-4chain, line 8: Some leftover ?

Makefile
pawellewandowski98: Added tools directory, I did not used it because it add another package (and we already have them a lot) and also another level in generated models name but If this is the right way so we can do it like that

api/components/models.yaml
pawellewandowski98: Divided models.go to `models.go requests.go responses.go`

api/configs/api.yaml
pawellewandowski98: Yes, removed

PR #877, opened 2025-01-29
feat(SPV-1125) finalize the story for op_return recording

chris-4chain:
# Pull Request Checklist


mgosek-4chain (approved):

actions/v2/operations/search.go
```diff
@@ -12,14 +12,14 @@ import (
 )
 
 func search(c *gin.Context, userContext *reqctx.UserContext) {
+	logger := reqctx.Logger(c)
+
 	userID, err := userContext.ShouldGetUserID()
 	if err != nil {
-		spverrors.ErrorResponse(c, spverrors.ErrCannotBindRequest, reqctx.Logger(c))
+		spverrors.AbortWithErrorResponse(c, err, logger)
```
dorzepowski, line 19: ### question / curiosity
What's the difference?

actions/v2/transactions/outlines_record_test.go
```diff
@@ -113,6 +114,35 @@ func TestOutlinesRecordOpReturn(t *testing.T) {
...
+		// and:
+		client := given.HttpClient().ForUser()
+
+		// when:
+		res, _ := client.R().
+			SetPathParams(map[string]string{
+				"txID": txSpec.ID(),
+				"vout": "0",
+			}).
+			Get("/api/v2/data/{txID}-{vout}")
```
dorzepowski, line 134: ### minor
Maybe you could just use outpoint.String() instead of 2 params?

engine/v2/data/interface.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"context"
+
+	"github.com/bitcoin-sv/spv-wallet/engine/v2/data/datamodels"
+	"github.com/bitcoin-sv/spv-wallet/models/bsv"
+)
+
+// Repo is the interface that wraps the basic operations with data.
+type Repo interface {
+	FindForUser(ctx context.Context, outpoint bsv.Outpoint, userID string) (*datamodels.Data, error)
```
dorzepowski, line 12: Maybe it should be ID instead of outpoint, and should be simply a string?
WDYT?
Because now it's that low level way of how we build an ID is exposed to all layers

actions/v2/data/routes.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"github.com/bitcoin-sv/spv-wallet/server/handlers"
+	routes "github.com/bitcoin-sv/spv-wallet/server/handlers"
+)
+
+// RegisterRoutes creates the specific package routes
+func RegisterRoutes(handlersManager *routes.Manager) {
+	group := handlersManager.Group(routes.GroupAPIV2, "/data")
+	group.GET("/:id", handlers.AsUser(get))
```
dorzepowski, line 11: We should have tests for this endpoint.
You didn't put test any error cases in outlines_record_test, like:
- login as admin
- login as anonymous
- wrong id

And IMHO we should have some easy way to test some happy path because otherwise we need to find the test in transactions and run whole flow just to find out how this endpoint is working or debug it.

dorzepowski (changes requested):

actions/v2/operations/search.go
chris-4chain: I changed it only for consistency reasons because I saw haw we do similar things for other endpoints.
Abort stops the middleware chain and writes the status and response right away.

actions/v2/transactions/outlines_record_test.go
chris-4chain: ✅ Done

actions/v2/data/routes.go
chris-4chain: ✅ Done
- added error cases
- added happy path along with `Faucet::StoreData` fixture method.

dorzepowski (approved):

PR #882, opened 2025-01-31
feat(SPV-1429): paymailDomain validation on domain-services level & error mapping for http response

chris-4chain:
# Pull Request Checklist


engine/v2/paymails/paymails_service.go
```diff
@@ -15,18 +16,24 @@ import (
 type Service struct {
 	paymailsRepo PaymailRepo
 	usersService UsersService
+	config       *config.AppConfig
```
mgosek-4chain, line 19: A domain service should focus solely on business logic and domain rules, without being directly tied to technical concerns like persistence, networking, or other infrastructure elements. In this case, we’re breaking that principle by introducing a dependency on the configuration.

engine/spverrors/map_error_response.go
```diff
@@ -0,0 +1,101 @@
...
+// ResponseMapper provides a fluent API for mapping errors to structured HTTP responses.
+// It allows conditional error handling with predefined responses while ensuring proper logging.
+type ResponseMapper interface {
+	// If checks uses errors.Is to match the error to the provided one.
+	If(errToMatch error) OnMatch
+	// Else sets the default error to return if no match was found.
+	Else(errToReturn models.ExtendedError)
+	// Finalize logs the base error and sends the response to the client.
+	Finalize()
+}
```
mgosek-4chain, line 20: I’m not entirely convinced we need to translate concepts from different technologies. Why not stick with straightforward functionality, like a simple error-handling function that handles errors based on their type for each handler?

engine/v2/users/users_service.go
```diff
@@ -4,24 +4,32 @@ import (
...
 	primitives "github.com/bitcoin-sv/go-sdk/primitives/ec"
+	"github.com/bitcoin-sv/spv-wallet/config"
 	"github.com/bitcoin-sv/spv-wallet/engine/spverrors"
 	"github.com/bitcoin-sv/spv-wallet/engine/v2/users/usersmodels"
 )
 
 // Service is a user domain service
 type Service struct {
 	usersRepo UserRepo
+	config    *config.AppConfig
```
mgosek-4chain, line 15: Same thing mentioned in the pervious comment.

config/config.go
```diff
@@ -196,7 +196,7 @@ type PaymailConfig struct {
 func (p *PaymailConfig) CheckDomain(domain string) error {
 	if p.DomainValidationEnabled {
 		if !slices.Contains(p.Domains, domain) {
```
wregulski, line 202: I see that it wasn't added in this specific PR, but I am thinking if extending this config.go file with new "utils" functions working on it's struct is a good idea.

I would consider to move those "utils" method under config package but separated file. 
Maybe not in this specific task, but I wanted to highlight it.

engine/v2/paymails/paymails_service.go
chris-4chain: I've switched to other approach - created `DomainChecker` interface in the service level and then make `config` struct to fullfill the interface requirement. 

Check it out, how it works right now 🤞

engine/spverrors/map_error_response.go
chris-4chain: This is my proposition how we can define (in fluent API) a mapping from "internal errors" to errors which are returned to user. 

Please give it a chance and take a look how it works 🤞 
Or, you could suggest me how can I solve this differently (in my particular case).

mgosek-4chain (approved):

engine/v2/paymails/paymails_service.go
chris-4chain: As we discussed - I've reverted to the `config` approach

wregulski (approved):

PR #887, opened 2025-02-04
feat(SPV-1398): select inputs for outline

dorzepowski:
# Pull Request Checklist


mgosek-4chain (approved):

actions/v2/transactions/outlines.go
```diff
@@ -31,7 +38,14 @@ func transactionOutlines(c *gin.Context, userCtx *reqctx.UserContext) {
...
 	}
 
-	txOutline, err := reqctx.Engine(c).TransactionOutlinesService().CreateBEEF(c, spec)
+	var txOutline *outlines.Transaction
+	switch format {
+	case bsv.TxHexFormatRAW:
+		txOutline, err = reqctx.Engine(c).TransactionOutlinesService().CreateRawTx(c, spec)
+	case bsv.TxHexFormatBEEF:
+		txOutline, err = reqctx.Engine(c).TransactionOutlinesService().CreateBEEF(c, spec)
+	}
```
chris-4chain, line 47: Shouldn't it be done in domain service?
Like `CreateOutline` method (which internally does the `switch-case`?

actions/v2/transactions/outlines.go
```diff
@@ -40,3 +54,16 @@ func transactionOutlines(c *gin.Context, userCtx *reqctx.UserContext) {
...
+
+func getOutlineTransactionFormat(c *gin.Context) (bsv.TxHexFormat, error) {
+	queryFormat, _ := c.GetQuery("format")
+	if queryFormat == "" {
+		return bsv.TxHexFormatBEEF, nil
+	}
+
+	format, err := bsv.ParseTxHexFormat(queryFormat)
+	if err == nil {
+		return format, spverrors.Wrapf(err, "invalid transaction format [%s] provided for transaction outline", queryFormat)
```
chris-4chain, line 66: Consider returning here a `spverror` with propper error code and http code 
- (because it is returned right to the user)

engine/testabilities/fixture_faucet.go
```diff
@@ -43,7 +43,12 @@ func (f *faucetFixture) TopUp(satoshis bsv.Satoshis) fixtures.GivenTXSpec {
 					bsv.Outpoint{TxID: txSpec.ID(), Vout: 0},
 					f.user.ID(),
 					satoshis,
-					nil,
+					bsv.CustomInstructions{
+						{
+							Type:        "sign",
+							Instruction: "P2PKH",
+						},
+					},
```
chris-4chain, line 51: My initial idea was that, `TopUp` method returns also `CustomInstructions`.
Just saying 🤷

engine/v2/transaction/outlines/inputs_specifications.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/bitcoin-sv/spv-wallet/engine/v2/transaction"
+	txerrors "github.com/bitcoin-sv/spv-wallet/engine/v2/transaction/errors"
+)
+
+// InputsSpec are representing a client specification for inputs part of the transaction.
+type InputsSpec struct {
+}
+
+func (s *InputsSpec) evaluate(ctx *evaluationContext, outputs annotatedOutputs) (annotatedInputs, error) {
+	outs, _ := outputs.splitIntoTransactionOutputsAndAnnotations()
```
chris-4chain, line 16: I would suggest to make those methods for `annotatedOutputs` type:

```golang
func (a annotatedOutputs) splitIntoTransactionOutputsAndAnnotations() ([]*sdk.TransactionOutput, transaction.OutputsAnnotations) {
	return a.toTransactionOutputs(), a.toAnnotations()
}

func (a annotatedOutputs) toTransactionOutputs() []*sdk.TransactionOutput {
	outputs := make([]*sdk.TransactionOutput, len(a))
	for i, out := range a {
		outputs[i] = out.TransactionOutput
	}
	return outputs
}

func (a annotatedOutputs) toAnnotations() transaction.OutputsAnnotations {
	annotations := make(transaction.OutputsAnnotations)
	for outputIndex, out := range a {
		if out.OutputAnnotation != nil {
			annotations[outputIndex] = out.OutputAnnotation
		}
	}
	return annotations
}
```

And then, use here `outputs.toTransactionOutputs`
This way we don't waste memory for "not-used" annotations.

engine/v2/transaction/outlines/inputs_specifications.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	utxos, err := ctx.UTXOSelector().Select(ctx, tx, ctx.UserID())
+	if err != nil {
+		return nil, spverrors.ErrInternal.Wrap(err)
+	}
+
+	if len(utxos) == 0 {
+		return nil, txerrors.ErrTxOutlineInsufficientFunds
+	}
```
chris-4chain, line 29: Do we do any checking if selected `utxos` actually fulfills the required satoshi amount?

chris-4chain (approved):

actions/v2/transactions/outlines.go
dorzepowski: IMHO it's kind of formatting decision driven by user that's why I see it at this level.

engine/v2/transaction/outlines/inputs_specifications.go
dorzepowski: I'm not sure if it answers you'r question, but [there are tests for selector to work properly](https://github.com/bitcoin-sv/spv-wallet/blob/9f9546d53050af37361e3a36d412a1b7070b1a7a/engine/v2/transaction/outlines/utxo/internal/sql/selector_test.go#L19)

Maybe we should make some verification regarding this when the outline is recorded because after signing transaction we could actually finish with not enough funds for it.

engine/testabilities/fixture_faucet.go
dorzepowski: Yes, and we agreed then, that you would fill it with something like this, just to not add all those additional derivation,
but you end up with nil.
Just saying 🤷

engine/v2/transaction/outlines/inputs_specifications.go
dorzepowski: Good point will do that (but in the next PR)

actions/v2/transactions/outlines.go
dorzepowski: bsv.ParseTxHexFormat is returning spverror (proper one IMHO) that's why I only wrapping it here

PR #905, opened 2025-02-12
feat(SPV-1465): add source transaction inputs HEX storage. 

mgosek-4chain:
# Pull Request Checklist

- [X] 📖 I created my PR using provided  : [CODE_STANDARDS](https://github.com/bitcoin-sv/spv-wallet/blob/main/.github/CODE_STANDARDS.md)
- [X] 📖 I have read the short Code of Conduct: [CODE_OF_CONDUCT](https://github.com/bitcoin-sv/spv-wallet/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] 🏠 I tested my changes locally.
- [X] ✅ I have provided tests for my changes.
- [X] 📝 I have used conventional commits.
- [X] 📗 I have updated any related documentation.
- [X] 💾 PR was issued based on the Github or Jira issue.


engine/v2/transaction/record/tx_flow.go
```diff
@@ -162,5 +163,24 @@ func (f *txFlow) broadcast() error {
 }
 
 func (f *txFlow) save() error {
+	isRawTx, err := f.service.transactions.HasTransactionInputSources(f.ctx, f.tx.Inputs...)
+	if err != nil {
+		return spverrors.Wrapf(err, "database query failed to check input source transactions for transaction %s", f.txID)
+	}
+
+	switch {
```
dorzepowski, line 171: ### matter of taste

I'm not sure if this should be really a swich - it seems like a simple if else case
but matter of taste

engine/v2/transaction/record/tx_flow.go
```diff
@@ -162,5 +163,24 @@ func (f *txFlow) broadcast() error {
...
+	case isRawTx:
+		f.txRow.SetRawHex(f.tx.Hex())
+		f.txRow.SetTransactionInputSources()
```
dorzepowski, line 174: it seems that SetRawHex should make setTransactionInputSources internally 
(I mean that I don't see a reason to split this into 2 separate methods.)

engine/v2/database/repository/operations.go
```diff
@@ -98,27 +98,38 @@ func mapOperations(operations iter.Seq[*txmodels.NewOperation]) []database.Opera
...
-		ID:       transaction.Transaction.ID,
-		TxStatus: string(transaction.Transaction.TxStatus),
+		ID:       operation.Transaction.ID,
+		TxStatus: string(operation.Transaction.TxStatus),
+		BeefHex:  operation.Transaction.BEEFHex(),
+		RawHex:   operation.Transaction.RawHex(),
 	}
 
-	for _, input := range transaction.Transaction.Inputs {
+	if !operation.Transaction.HasTransactionInputSources() {
```
dorzepowski, line 109: do we really need this? 
If the input sources list is empty, the loop won't run, so this check seem to be ambiguous.

actions/paymailserver/incoming_paymail_tx_test.go
```diff
@@ -332,6 +332,32 @@ func TestIncomingPaymailBeef(t *testing.T) {
...
+			}`
+
+		// when:
+		res, _ := recipientClient.R().
+			SetHeader("Content-Type", "application/json").
+			SetBody(requestBody).
+			Post("/api/v2/transactions/outlines")
+
+		//  then:
+		thenResponse := testabilities.NewTransactionsEndpointAssertions(t, given).Response(res)
```
dorzepowski, line 357: ### just informative comment

I don't like this change. This should be done somehow differently, so you can use the same `then` here (which should be a dedicated `testability assertion` for those tests.
But I don't know how it could be done, and it's possible that actually it can't be done in Go because there is not inheritance but only wrapping.

engine/v2/transaction/txmodels/new_transaction.go
```diff
@@ -1,19 +1,69 @@
...
+func (t *NewTransaction) SetTransactionInputSources() {
+	slice := make([]TransactionInputSource, 0, len(t.Outputs))
+	for _, o := range t.Outputs {
+		slice = append(slice, TransactionInputSource{
+			TxID:       t.ID,
+			SourceTxID: o.TxID,
+		})
+	}
+
+	t.transactionInputSources = slice
```
dorzepowski, line 64: ### example alternative
```suggestion
	t.transactionInputSources = lo.Map(t.Outputs, func(item NewOutput, index int) TransactionInputSource {
		return TransactionInputSource{
			TxID:       t.ID,
			SourceTxID: item.TxID,
		}
	})
```

engine/v2/database/repository/transactions.go
```diff
@@ -0,0 +1,130 @@
...
+	// Fetch input sources in batches
+	total, err := t.queryInputSourcesBatch(ctx, sourceTXIDs, visited)
+	if err != nil {
+		return nil, spverrors.Wrapf(err, "failed to query input sources batches")
+	}
+
+	// Convert to TxQueryResultSlice
+	var slice beef.TxQueryResultSlice
+	for _, record := range total {
+		slice = append(slice, record.ToTxQueryResult())
```
dorzepowski, line 60: It would be better to select into TxQueryResult struct, to limit the data that are queried. I believe that this is possible with Gorm.
But maybe it could be done in separate task because I also suggest to make it with single db query, so maybe it could solve both "comments" at once.

engine/v2/database/repository/transactions.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	return slice, nil
+}
+
+// queryInputSourcesBatch retrieves transactions in batches to optimize database performance.
```
dorzepowski, line 65: don't add comments to private methods, because if the implementation will change, then comments can be outdated.

engine/v2/database/repository/transactions.go
```diff
@@ -0,0 +1,130 @@
...
+		results = append(results, row)
+		if !row.HasBeefHex() {
+			for _, input := range row.SourceTxInputs {
+				nextBatch = append(nextBatch, input.SourceTxID)
+			}
+		}
+	}
+
+	// Recursively fetch the next batch
+	nextResults, err := t.queryInputSourcesBatch(ctx, nextBatch, visitedTransactions)
```
dorzepowski, line 109: Maybe lets add a JIRA task to try to query those things with single SQL query?
If you think that it's good idea, please create such task in some of the next sprints.

engine/v2/database/repository/transactions.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	// Filter out already visited transactions
+	filteredIDs := make([]string, 0, len(txIDs))
+	for _, txID := range txIDs {
+		if !visitedTransactions.isVisited(txID) {
+			visitedTransactions.recordVisited(txID)
+			filteredIDs = append(filteredIDs, txID)
+		}
+	}
```
dorzepowski, line 79: ### alternative suggestion
```suggestion
	filteredIDs := lo.Filter(txIDs, lox.MappingFn(visitedTransactions.isNotVisited))
	
	lo.ForEach(filteredIDs, func(txID string, _ int) {
		visitedTransactions.recordVisited(txID)
	})
```

engine/v2/database/repository/transactions.go
```diff
@@ -0,0 +1,130 @@
...
+		WithContext(ctx).
+		Preload("SourceTxInputs").
+		Where("id IN (?)", filteredIDs).
+		Find(&rows).Error
+	if err != nil {
+		return nil, spverrors.Wrapf(err, "failed to batch query transactions: %v", filteredIDs)
+	}
+
+	// Process results and collect next batch of transaction IDs
+	results := make([]database.TrackedTransaction, 0, len(rows))
```
dorzepowski, line 97: why we're copying those results?

engine/v2/database/repository/transactions.go
```diff
@@ -0,0 +1,130 @@
...
+	nextResults, err := t.queryInputSourcesBatch(ctx, nextBatch, visitedTransactions)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(results, nextResults...), nil
+}
+
+// visitedTransactions is a helper type that tracks visited transactions during recursive queries.
+type visitedTransactions map[string]bool
```
dorzepowski, line 118: Wouldn't it be more effective if we use `map[string]struct{}`

dorzepowski (changes requested):

engine/v2/transaction/record/tx_flow.go
mgosek-4chain: Fixed :)

engine/v2/transaction/record/tx_flow.go
mgosek-4chain: Agree & fixed :)

engine/v2/database/repository/transactions.go
mgosek-4chain: Fixed

actions/paymailserver/incoming_paymail_tx_test.go
mgosek-4chain: I see your point. I plan to replace this after merging [SPV-1059](https://github.com/bitcoin-sv/spv-wallet/tree/feat/SPV-1059-fee-and-change).

engine/v2/transaction/txmodels/new_transaction.go
mgosek-4chain: Fixed

engine/v2/database/repository/transactions.go
mgosek-4chain: The initial goal was to provide a basic implementation with the minimum necessary changes in this PR. However, as you mentioned, I’d like to focus on improving overall performance in a new task.

engine/v2/database/repository/transactions.go
mgosek-4chain: Done :)

engine/v2/database/repository/transactions.go
mgosek-4chain: Due to the known number of queried rows from the following code part: 

```
// Batch query transactions
	var rows []database.TrackedTransaction
	err := t.db.
		WithContext(ctx).
		Preload("SourceTxInputs").
		Where("id IN (?)", filteredIDs).
		Find(&rows).Error
	if err != nil {
		return nil, spverrors.Wrapf(err, "failed to batch query transactions: %v", filteredIDs)
	}
```

engine/v2/transaction/record/tx_flow.go
```diff
@@ -162,5 +163,20 @@ func (f *txFlow) broadcast() error {
...
+
+	if isRawTx {
+		f.txRow.SetRawHex(f.tx.Hex())
+	} else {
+		hex, err := f.tx.BEEFHex()
+		if err != nil {
+			return spverrors.Wrapf(err, "failed to generate BEEF hex for transaction %s", f.txID)
+		}
+		f.txRow.SetBEEFHex(hex)
+	}
```
chris-4chain, line 179: This could be done at the initial step (instead of final step 'save()').
- We do have every information at initial 
- There is no point to calculate all of the things and then return error in the end. 
- It's also kind of dangerous because if broadcast and/or notify p2p happens, the transaction starts to be real - so we should save it
- That's why, maybe it's better to make this check and hex assignment in `newTxFlow` function 🤔

engine/v2/transaction/record/tx_flow.go
```diff
@@ -162,5 +163,20 @@ func (f *txFlow) broadcast() error {
 }
 
 func (f *txFlow) save() error {
+	isRawTx, err := f.service.transactions.HasTransactionInputSources(f.ctx, f.tx.Inputs...)
+	if err != nil {
+		return spverrors.Wrapf(err, "database query failed to check input source transactions for transaction %s", f.txID)
+	}
```
chris-4chain, line 169: What if HandCash (or other) sends us "raw tx"? Maybe it's the matter for other PR - but, just thinking 🤔

chris-4chain (changes requested):

engine/v2/transaction/record/tx_flow.go
dorzepowski: Good catch, I agree.

engine/v2/transaction/record/tx_flow.go
```diff
@@ -39,6 +40,33 @@ func newTxFlow(ctx context.Context, service *Service, tx *trx.Transaction) *txFl
...
+	if err := f.setTxRowHex(); err != nil {
+		return nil, err
+	}
+
+	return f, nil
+}
+
+func (f *txFlow) setTxRowHex() error {
+	txID := f.tx.TxID().String()
+	isRawTx, err := f.service.transactions.HasTransactionInputSources(f.ctx, f.tx.Inputs...)
```
chris-4chain, line 53: To be honest the name `isRawTx` confuses me. 
The method which produces value of this variable is `HasTransactionInputSources` so the returned value says "at least one source transaction is in database". 

All I want to say is this is kind of crucial decision point and it's not obvious. 
Maybe with some comment 🤔 

like this:
```golang
knownSource, err := f.service.transactions.HasTransactionInputSources(f.ctx, f.tx.Inputs...)
	if err != nil {
		return spverrors.Wrapf(err, "database query failed to check input source transactions for transaction %s", txID)
	}

	if knownSource {
		// some description why if "at least one source transaction is in database" we set raw tx here
		f.txRow.SetRawHex(f.tx.Hex())
		return nil
	}
```

engine/v2/database/repository/transactions.go
```diff
@@ -0,0 +1,121 @@
...
+func (t *Transactions) queryInputSourcesBatch(ctx context.Context, txIDs []string, visitedTransactions visitedTransactions) ([]database.TrackedTransaction, error) {
+	if len(txIDs) == 0 {
+		return nil, nil
+	}
+
+	// Filter out already visited transactions
+	filteredIDs := lo.Filter(txIDs, lox.MappingFn(visitedTransactions.isNotVisited))
+	lo.ForEach(filteredIDs, func(txID string, _ int) {
+		visitedTransactions.recordVisited(txID)
+	})
```
chris-4chain, line 78: Just a thought.
For crucial points of our algorithms I would avoid usage of `lo`.
We've introduced the `lo` mainly for `mapping` purposes. 

I think, vanilla golang would be more efficient and "maybe" more familiar to all golang developers.

mgosek-4chain: @dorzepowski please check this out :)

engine/v2/transaction/record/tx_flow.go
mgosek-4chain: Agree. Maybe `hasAnySourceTx` 🤔 ?

engine/v2/database/repository/transactions.go
```diff
@@ -0,0 +1,121 @@
...
+
+// NewTransactions creates a new Transactions repository.
+// It initializes a database-backed service for querying and managing transactions.
+func NewTransactions(db *gorm.DB) *Transactions {
+	return &Transactions{db: db}
+}
+
+// HasTransactionInputSources checks if any of the given transaction inputs have a source transaction in the database.
+// It queries the database to determine if at least one of the provided input source transaction IDs exists.
+func (t *Transactions) HasTransactionInputSources(ctx context.Context, inputs ...*trx.TransactionInput) (bool, error) {
```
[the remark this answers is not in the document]
mgosek-4chain: Good point. Fixed 😸

engine/v2/transaction/record/tx_flow.go
chris-4chain: `HasAnyTransactionInputSources` would be nice :) 

But still, the `isRawTx` variable name confuses me 🤷

engine/v2/database/repository/transactions.go
```diff
@@ -0,0 +1,126 @@
...
+// If all of them are found, the transaction data can be serialized into Raw HEX format.
+// Otherwise, serialization should be done using the BEEFHex format.
+func (t *Transactions) HasTransactionInputSources(ctx context.Context, sourceTXIDs ...string) (bool, error) {
+	set := make(map[string]struct{})
+	for _, txID := range sourceTXIDs {
+		set[txID] = struct{}{}
+	}
+
+	keys := maps.Keys(set)
+	ids := slices.AppendSeq(make([]string, 0, len(set)), keys)
```
chris-4chain, line 35: **Minor**

```suggestion
	set := make(map[string]struct{})
	for _, txID := range sourceTXIDs {
		set[txID] = struct{}{}
	}
	keys := maps.Keys(set)
	uniqueIds := slices.AppendSeq(make([]string, 0, len(set)), keys)
```

chris-4chain (approved):

dorzepowski (approved):

PR #923, opened 2025-02-24
test(SPV-1478): helper tool for making manual tests for V2

dorzepowski:
# Pull Request Checklist


api/manualtests/adminapi/errorcases/admin_create_user_badreq_test.go
```diff
@@ -0,0 +1,87 @@
...
+			require.NoError(t, err)
+
+			user := manualtests.User{
+				Alias:     "badrequest",
+				Domain:    state.Domain,
+				Xpriv:     "xprv9s21ZrQH143K3TzwskUB1NW5iHx4EH7cxquXfwpFR5HrWe6HQrYvECJsj3sg1DJhWhwjtw5WdXwje8pkyuuvzJUingwo4f5BkD5dNubfNUn",
+				Xpub:      "xpub661MyMwAqRbcFx5Qyn1BNWSpGKnYdjqUL4q8ULDryQpqPSRRxPsAmzdMaLnoyUzLAQ5ukXgMZjYs5LfNfsPFwBoSwxChePB1DxKvyFz6F67",
+				PublicKey: "035dcb59eb7b5c5982ba6fbbccffbb2460f8daa07c5f9a21f2c2cf0845dcc6dda9",
+				ID:        "1QCibNnc8CK7bzTdupM91f4PKixhMqqJQw",
+			}
```
mgosek-4chain, line 74: The same definition appears in the previous test. Would it make sense to extract it into a constructor function for reuse? 🤔

api/manualtests/test_state.go
```diff
@@ -0,0 +1,325 @@
...
+	OldUsers              []User  `mapstructure:"zzz_old_users" yaml:"zzz_old_users"`
+	configFilePath        string
+	configFileJustCreated bool
+}
+
+func NewState() *State {
+	state := &State{
+		Domain:    notConfiguredPaymailDomain,
+		ServerURL: "http://localhost:3003",
+		AdminXpub: "xpub661MyMwAqRbcFgfmdkPgE2m5UjHXu9dj124DbaGLSjaqVESTWfCD4VuNmEbVPkbYLCkykwVZvmA8Pbf8884TQr1FgdG2nPoHR8aB36YdDQh",
```
mgosek-4chain, line 43: Would it make sense to use constants instead of hardcoding these values?

api/manualtests/test_state.go
```diff
@@ -0,0 +1,325 @@
...
+		return nil, errorx.Decorate(err, "failed to unlock outline hex")
+	}
+	return tx, nil
+}
+
+func (s *State) SaveUserDataOutpoint(txID string, vout int) error {
+	s.CurrentUser().AddDataOutpoint(txID, vout)
+	err := s.Save()
+	if err != nil {
+		return err
```
mgosek-4chain, line 322: Could you explain why errorx.Decorate isn’t applied consistently? Just curious about the reasoning behind it.

api/manualtests/test_state_client.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (s *State) authAsCurrentUser() client.RequestEditorFn {
+	return authWithKey(s.User.Xpub)
+}
+
+func (s *State) authAsUnknown() client.RequestEditorFn {
+	// unknownXpriv := "xprv9s21ZrQH143K3jw372AgTDppRcaFZZiikzCbZodX4CqzZM8SZdqkBgcmmp5DKtmaKFNcJq9AcpFd35X2oyKXJogmnziU5h7tb72qSFghNnA"
+	unknownXpub := "xpub661MyMwAqRbcGE1WD3hgpMmYyeQjy2Sa8D8CNC38cYNyS9Tb7B9zjUwFd5ZSzFYjKMWsNzh1sxb5AspdaSp31Zqz9CDM5GL74Eib22h25Pv"
+	return authWithKey(unknownXpub)
+}
```
mgosek-4chain, line 87: Would it make sense to add two separate functions — `authAsUnknownWithXPub` and `authAsUnknownWithXPriv` — for clarity and better separation of concerns?

mgosek-4chain (approved):

chris-4chain (approved):

api/manualtests/test_state.go
dorzepowski: I don't see a reason for that.
Bear in mind it's just a tool for manual testing, not for automatic testing,
and for me if we don't reuse something, then it's good enough to have it here.

api/manualtests/test_state.go
dorzepowski: My thought was that, if some error is returned from method called on the same receiver (the same struct) then adding additional decoration, probably won't give any additional benefits.

And again, it's just a tool for manual testing - it isn't production ready nor it will be in the future. So it can skip some good practice - it just need to work.

api/manualtests/test_state_client.go
dorzepowski: Unfortunately there is no auth module, therefore the only authentication we have here is with xpub - 
It means that on tested spv-wallet you need to have signing disabled.
I put the xpriv here, just for the future if we will add some authorization with signing to those tests, so we don't need to change the key pair, because we already have the xpriv noted down.

api/manualtests/adminapi/errorcases/admin_create_user_badreq_test.go
dorzepowski: done

PR #926, opened 2025-02-25
feat(SPV-1555) record change output

chris-4chain:
# Pull Request Checklist


engine/tester/fixtures/unlocking_resolver.go
```diff
@@ -0,0 +1,43 @@
...
+	return false, nil
+}
+
+func (un *UnlockingTemplateResolver) Finalize(acc *custominstructions.Accumulator[primitives.PrivateKey]) error {
+	if un.Template == nil {
+		// this is implicit "Sign" if there was no "Sign" instruction in provided custom instructions
+		_, err := un.Sign(acc, "P2PKH")
+		if err != nil {
+			return err
+		}
```
mgosek-4chain, line 40: Why not call panic, as done in the previous cases, instead of returning an error here? 🤔

engine/v2/custominstructions/locking_resolver.go
```diff
@@ -0,0 +1,79 @@
...
+	return false, nil
+}
+
+// Finalize ensures that the address is resolved.
+func (ar *AddressResolver) Finalize(acc *Accumulator[primitives.PublicKey]) error {
+	if ar.Address == nil {
+		// this is implicit "Sign" if there was no "Sign" instruction in provided custom instructions
+		_, err := ar.Sign(acc, "P2PKH")
+		if err != nil {
+			return err
```
mgosek-4chain, line 52: Could you explain why the error isn't being wrapped here? 🤔

mgosek-4chain (approved):

engine/v2/custominstructions/locking_resolver.go
chris-4chain: Because the error is already wrapped and produce in the same package in `Sign` method. 

But I can add additional "context" for the error and wrap it 🙈

engine/v2/keys/type42/destination.go
```diff
@@ -25,7 +25,7 @@ func newDestinationWithReference(pubKey *primitives.PublicKey, referenceID strin
 	}
 
 	dst.DerivationKey = fmt.Sprintf("1-destination-%s", referenceID)
-	dst.PubKey, err = derive(pubKey, dst.DerivationKey)
+	dst.PubKey, err = Derive(pubKey, dst.DerivationKey)
 	if err != nil {
 		return dst, err
```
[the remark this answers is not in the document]
chris-4chain: I can add additional "context" for the error and wrap it 🙈

engine/v2/transaction/outlines/output_specifications.go
```diff
@@ -91,7 +92,11 @@ func (a annotatedOutputs) toAnnotations() transaction.OutputsAnnotations {
 	annotations := make(transaction.OutputsAnnotations)
 	for outputIndex, out := range a {
 		if out.OutputAnnotation != nil {
-			annotations[outputIndex] = out.OutputAnnotation
+			vout32, err := conv.IntToUint32(outputIndex)
+			if err != nil {
+				panic(spverrors.Wrapf(err, "failed to convert output index %d to uint32", outputIndex))
```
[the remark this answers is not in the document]
chris-4chain: Yes.
Generally this error "almost never" happen. 
I could skip this gosec linter error as well. But I decided to call `panic`.

engine/v2/custominstructions/interpreter_test.go
```diff
@@ -0,0 +1,79 @@
...
+		})
+	}
+}
+
+func makePubKey(t *testing.T, pubDERHex string) *primitives.PublicKey {
+	t.Helper()
+	pk, err := primitives.PublicKeyFromString(pubDERHex)
+	if err != nil {
+		t.Fatalf("failed to create public key: %s", err)
+	}
```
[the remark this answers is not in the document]
chris-4chain: ✅ Done

engine/tester/fixtures/unlocking_resolver.go
chris-4chain: ✅ Done

engine/v2/custominstructions/locking_resolver.go
chris-4chain: ✅ Done

engine/v2/keys/type42/destination.go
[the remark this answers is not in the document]
chris-4chain: ✅ Done

engine/v2/custominstructions/locking_resolver.go
mgosek-4chain: It's up to you, I was just curious 😄

wregulski (approved):

engine/v2/transaction/record/custom_outputs.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (c *customOutputsResolver) resolveAddress(address string, vout uint32) {
+	c.txAddresses.remove(address, vout)
+}
+
+func (c *customOutputsResolver) annotatedOutputs() iter.Seq2[string, txmodels.NewOutput] {
+	return func(yield func(address string, output txmodels.NewOutput) bool) {
+		for vout, annotation := range c.annotations {
+			if annotation.Bucket != bucket.BSV || annotation.CustomInstructions == nil {
```
dorzepowski, line 46: maybe it's enough to check if it has CustomInstructions?

engine/v2/custominstructions/interpreter.go
```diff
@@ -0,0 +1,50 @@
...
+func (p *Interpreter[I, TKey]) Process(key *TKey, instructions bsv.CustomInstructions) (I, error) {
+	acc := &Accumulator[TKey]{
+		Key: key,
+	}
+	var err error
+	var proceed bool
+	for _, instruction := range instructions {
+		switch instruction.Type {
+		case Type42:
+			proceed, err = p.resolver.Type42(acc, instruction.Instruction)
```
dorzepowski, line 30: ### minor
I just wonder if accumulator shouldn't be a part of resolver, so we don't need to pass it to each and every method

engine/v2/custominstructions/interpreter.go
```diff
@@ -0,0 +1,50 @@
...
+			break
+		}
+	}
+
+	err = p.resolver.Finalize(acc)
+	if err != nil {
+		return p.resolver, errors.ErrFinalizingCustomInstructions.Wrap(err)
+	}
+
+	return p.resolver, nil
```
dorzepowski, line 49: ### minor
And I wonder if we shouldn't return Accumulator with locking / unlocking script instead of resolver.

engine/v2/transaction/record/custom_outputs.go
```diff
@@ -0,0 +1,91 @@
...
+			}
+
+			interpreted, err := custominstructions.NewAddressInterpreter().
+				Process(userPubKey, *annotation.CustomInstructions)
+			if err != nil {
+				c.err = spverrors.Wrapf(err, "failed to derive address from custom instructions for user %s", c.userID)
+				break
+			}
+
+			_, ok := c.txAddresses[interpreted.Address.AddressString]
```
dorzepowski, line 63: I totally don't understand, why we're storing all addresses and why we're looking for them.

I mean: we've got annotation that is strictly saying:
> Output X: will be unlocked with this custom instructions
 
But then, we're checking if those custom instructions will unlock any of Outputs, not only X

engine/v2/transaction/record/custom_outputs.go
```diff
@@ -0,0 +1,91 @@
...
+			if !ok {
+				c.err = spverrors.Newf("address derived from custom instructions doesn't match any of the addresses in the locking scripts")
+				break
+			}
+
+			if ok := c.txAddresses.contains(interpreted.Address.AddressString, vout); !ok {
```
dorzepowski, line 69: why we don't just use the tx outputs, and if `output[vout]` contains address?

engine/v2/custominstructions/interpreter.go
```diff
@@ -0,0 +1,50 @@
...
+// Process processes custom instructions for a given key.
+func (p *Interpreter[I, TKey]) Process(key *TKey, instructions bsv.CustomInstructions) (I, error) {
+	acc := &Accumulator[TKey]{
+		Key: key,
+	}
+	var err error
+	var proceed bool
+	for _, instruction := range instructions {
+		switch instruction.Type {
+		case Type42:
```
dorzepowski, line 29: ### minor

Maybe 
because we're not prepared for that, 
we should remove the option to pass ANY custom instructions string in openapi and just leave the spv-wallet custom instructions type, WDYT?

engine/v2/custominstructions/types.go
```diff
@@ -0,0 +1,7 @@
+package custominstructions
+
+// Types for custom instructions
+const (
+	Type42 string = "type42"
+	Sign   string = "sign"
```
dorzepowski, line 6: I believe that we should specify type enum in openapi docs also

engine/v2/custominstructions/errors/errors.go
```diff
@@ -0,0 +1,45 @@
...
+	Code:       "error-custom-instructions-derivation-failed",
+	Message:    "Failed to derive type42 public key for given instruction",
+	StatusCode: 500,
+}
+
+// ErrUnknownInstructionType is returned when an unknown instruction type is encountered
+var ErrUnknownInstructionType = models.SPVError{
+	Code:       "error-unknown-instruction-type",
+	Message:    "Unknown instruction type",
+	StatusCode: 500,
```
dorzepowski, line 30:
```suggestion
	StatusCode: 400,
```

engine/v2/custominstructions/errors/errors.go
```diff
@@ -0,0 +1,45 @@
+package errors
+
+import "github.com/bitcoin-sv/spv-wallet/models"
+
+// ErrProcessingCustomInstructions is returned when custom instructions cannot be processed
+var ErrProcessingCustomInstructions = models.SPVError{
+	Code:       "error-custom-instructions-processing",
+	Message:    "Failed to process custom instructions",
+	StatusCode: 500,
```
dorzepowski, line 9:
```suggestion
	StatusCode: 422,
```

engine/v2/custominstructions/errors/errors.go
```diff
@@ -0,0 +1,45 @@
...
+	Code:       "error-custom-instructions-finalize",
+	Message:    "Failed to finalize processing custom instructions",
+	StatusCode: 500,
+}
+
+// ErrType42DerivationFailed is returned when a type42 public key cannot be derived
+var ErrType42DerivationFailed = models.SPVError{
+	Code:       "error-custom-instructions-derivation-failed",
+	Message:    "Failed to derive type42 public key for given instruction",
+	StatusCode: 500,
```
dorzepowski, line 23: ### not really important

It's wrapped into other error so it's not really important but:

```suggestion
	StatusCode: 422,
```

engine/v2/custominstructions/errors/errors.go
```diff
@@ -0,0 +1,45 @@
...
+	Code:       "error-unknown-instruction-type",
+	Message:    "Unknown instruction type",
+	StatusCode: 500,
+}
+
+// ErrGettingAddressFromPublicKey is returned when an address cannot be derived from a public key
+var ErrGettingAddressFromPublicKey = models.SPVError{
+	Code:       "error-getting-address-from-public-key",
+	Message:    "Failed to get address from public key",
+	StatusCode: 500,
```
dorzepowski, line 37: ### not really important

It's wrapped into other error so it's not really important but:

```suggestion
	StatusCode: 422,
```

engine/v2/custominstructions/errors/errors.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// ErrGettingLockingScript is returned when a locking script cannot be derived from an address
+var ErrGettingLockingScript = models.SPVError{
+	Code:       "error-getting-locking-script",
+	Message:    "Failed to get locking script from address",
+	StatusCode: 500,
```
dorzepowski, line 44: ### not really important

It's wrapped into other error so it's not really important but:

```suggestion
	StatusCode: 422,
```

engine/v2/transaction/record/custom_outputs.go
chris-4chain: I would leave it as is - I see it as more safe approach.
And when we have other use cases, we will adjust it 🤞

engine/v2/custominstructions/types.go
chris-4chain: Good point 👍 

I will do it in SPV-1552 (standarize spv-wallet custom instructions)
I didn't want to make too much changes in this PR

engine/v2/custominstructions/interpreter.go
chris-4chain: Let's discuss this by the way of future task: SPV-1552 (standarize spv-wallet custom instructions)

For now I would leave it as is.

engine/v2/transaction/record/custom_outputs.go
chris-4chain: We already did the address-extraction from outputs - because it is needed for `createUTXOsForTrackedAddresses` anyway.

engine/v2/transaction/record/custom_outputs.go
dorzepowski: I don't understand what are we trying to achieve with this check?
how it is making this more safe for us?

engine/v2/transaction/record/custom_outputs.go
dorzepowski: Ok, despite this address extaction,
why are we doing double check
first if there is this address at all
second if it is in provided vout

IMHO we should check only if it is in that particular output, and that's all.

dorzepowski (approved):

PR #928, opened 2025-02-25
fix(SPV-1561): cannot make new transaction on postgres

dorzepowski:
# Pull Request Checklist


mgosek-4chain (approved):

engine/v2/database/repository/outputs.go
```diff
@@ -42,9 +43,12 @@ func (o *Outputs) FindByOutpoints(ctx context.Context, outpoints iter.Seq[bsv.Ou
...
 		return txmodels.TrackedOutput{
-			TxID:       output.TxID,
-			Vout:       output.Vout,
-			SpendingTX: output.SpendingTX,
+			TxID: output.TxID,
+			Vout: output.Vout,
+			// Because the field is char(64) in the database, it will be padded with spaces.
+			// For some reason the value is padded only on postgres,
+			// so if changing, make sure to check it on all databases.
+			SpendingTX: strings.TrimLeft(output.SpendingTX, " "),
```
chris-4chain, line 51: Consider this (as a hook):

```golang
// AfterFind is a GORM hook that is called after retrieving the record from the database.
func (o *TrackedOutput) AfterFind(tx *gorm.DB) (err error) {
	// Trim left spaces from SpendingTX
	// Because the field is char(64) in the database, it will be padded with spaces.
	// For some reason the value is padded only on postgres,
	// so if changing, make sure to check it on all databases.
	o.SpendingTX = strings.TrimLeft(o.SpendingTX, " ")
	return nil
}
```
This way, if we have other functions which retrieve the TrackedOutput, the "trim-thing" will be done. 

I haven't tested it 🙈 ... 🤞 it will work xD

chris-4chain: ✅ Applied

chris-4chain (approved):

PR #933, opened 2025-02-26
Feat(SPV-1425): add validation for paymail

dzolt-4chain:
# Pull Request Checklist


ac4ch (approved):

actions/v2/admin/users/create_test.go
```diff
@@ -115,6 +115,18 @@ func TestCreateUserWithPaymail(t *testing.T) {
...
 	publicName := "Test User"
 
+	// and:
+	secondUserCandidate := fixtures.User{
+		PrivKey: "xprv9s21ZrQH143K4Tf1hf7ouMiagMH4JKvE6E2SY8Su55Y6aFi9AfQibzx7i79g1NJkLQbRY4FjKgvpddtYXoD7dA2KbGjHdHcxXVqtd687KrK",
+		Paymails: []fixtures.Paymail{
+			"second@" + fixtures.PaymailDomain,
+		},
+	}
+	secondPublicKey := secondUserCandidate.PublicKey().ToDERHex()
```
chris-4chain, line 125: To be honest, I don't like this approach witch "second-sth". I guess, that it was needed because you cannot test the cases with wrong url/public name. 

Moreover this was intended to be two-steps-test (1. Create & 2. Get - as extended "assertion")
You're breaking this approach. 

I would advocate for doing separate test(s) with your new cases.

actions/v2/admin/users/paymail_add.go
```diff
@@ -32,8 +33,8 @@ func (s *APIAdminUsers) AddPaymailToUser(c *gin.Context, id string) {
 	newPaymail := &paymailsmodels.NewPaymail{
 		Alias:      alias,
 		Domain:     domain,
-		PublicName: requestBody.PublicName,
-		Avatar:     requestBody.AvatarURL,
+		PublicName: lo.IfF(requestBody.PublicName != nil, func() string { return *requestBody.PublicName }).Else(""),
+		Avatar:     lo.IfF(requestBody.AvatarURL != nil, func() string { return *requestBody.AvatarURL }).Else(""),
```
chris-4chain, line 37: Hmmm 🤔 

I suppose, this block of code is ready to be moved to `internal/mapping`.
I mean, this code:
```golang
newUser := &usersmodels.NewUser{
		PublicKey: requestBody.PublicKey,
	}
	if requestBody.PaymailDefined() {
		alias, domain, err := parsePaymail(requestBody.PaymailRequestBody())
		if err != nil {
			spverrors.ErrorResponse(c, err, s.logger)
			return
		}

		newUser.Paymail = &usersmodels.NewPaymail{
			Alias:  alias,
			Domain: domain,

			PublicName: lo.IfF(requestBody.Paymail.PublicName != nil, func() string { return *requestBody.Paymail.PublicName }).Else(""),
			Avatar:     lo.IfF(requestBody.Paymail.AvatarURL != nil, func() string { return *requestBody.Paymail.AvatarURL }).Else(""),
		}
	}
```

Could you do this?

.... and also for `paymail_add.go`.

engine/v2/paymails/paymails_service.go
```diff
@@ -40,6 +40,13 @@ func (s *Service) Create(ctx context.Context, newPaymail *paymailsmodels.NewPaym
 		return nil, spverrors.Newf("user does not exist")
 	}
 
+	if isValid := newPaymail.CheckAvatarURL(); !isValid {
+		return nil, spverrors.Newf("invalid avatar url during paymail creation")
```
chris-4chain, line 44: This should be defined `paymailerrors` package with 500 http code. 
And then mapped to 422 error in `/actions/.../create.go` & `/actions/.../paymail_add.go`.
Similarly to: https://github.com/bitcoin-sv/spv-wallet/blob/5f2b29a72994c4cb61ea516cf2bc5511a361b671/actions/v2/admin/users/create.go#L52-L57

engine/v2/paymails/paymailsmodels/paymail.go
```diff
@@ -1,6 +1,11 @@
 package paymailsmodels
 
-import "time"
+import (
+	"regexp"
+	"time"
+)
+
+var explicitHTTPURLRegex = regexp.MustCompile(`^https?:\/\/(?:www\.)?[a-zA-Z0-9-]+(?:\.[a-zA-Z]{2,})+(?:\/[^\s]*)?`)
```
chris-4chain, line 8: Don't we allow `http` ?

engine/v2/users/users_service.go
```diff
@@ -31,6 +31,12 @@ func (s *Service) Create(ctx context.Context, newUser *usersmodels.NewUser) (*us
 		if err := s.config.Paymail.CheckDomain(newUser.Paymail.Domain); err != nil {
 			return nil, spverrors.Wrapf(err, "invalid domain during user creation")
 		}
+		if isValid := newUser.Paymail.CheckAvatarURL(); !isValid {
+			return nil, spverrors.Newf("invalid avatar url during user creation")
+		}
```
chris-4chain, line 36: You could return here the same error from `paymailerrors`, as described above ☝️

engine/v2/users/usersmodels/paymail.go
```diff
@@ -26,3 +31,8 @@ type NewPaymail struct {
 	Avatar     string
 	UserID     string
 }
+
+// CheckAvatarURL checks if avatar is either empty string or a proper url link
+func (np *NewPaymail) CheckAvatarURL() bool {
+	return np.Avatar == "" || explicitHTTPURLRegex.MatchString(np.Avatar)
+}
```
chris-4chain, line 38: Maybe it should be somehow refactored, because it is exactly the same code as in `paymailmodels`. 
Maybe this (`usermodels.paymail.go`) could be removed and used the one from `paymailmodels` 🤔 

I'm aware that some "circular dependency" issue can occur - but please try to think about it :)

chris-4chain (changes requested):

engine/v2/paymails/paymails_service.go
```diff
@@ -40,6 +40,13 @@ func (s *Service) Create(ctx context.Context, newPaymail *paymailsmodels.NewPaym
 		return nil, spverrors.Newf("user does not exist")
 	}
 
+	if isValid := newPaymail.CheckAvatarURL(); !isValid {
```
dorzepowski, line 43: why not:
```go
err := newPaymail.ValidateAvatar()
if err != nil {
    return nil, spverrors.Wrapf(err, "invalid avatar url")

}

```

engine/v2/paymails/paymailsmodels/paymail.go
```diff
@@ -26,3 +31,8 @@ type NewPaymail struct {
 	Avatar     string
 	UserID     string
 }
+
+// CheckAvatarURL checks if avatar is either empty string or a proper url link
+func (np *NewPaymail) CheckAvatarURL() bool {
+	return np.Avatar == "" || explicitHTTPURLRegex.MatchString(np.Avatar)
```
dorzepowski, line 37: maybe instead of some hard to understand regex use:
```go
		URL, err := url.Parse(avatarURL)
		if err != nil {
			return ErrInvalidAvatarURL.Wrap(err)
		}

		if URL.Scheme != "http" || URL.Scheme != "https" {
           return ErrInvalidAvatarURL.Wrap(spverrors.Newf("avatarURL should have http(s) scheme"))
        }
```

actions/v2/admin/users/paymail_add.go
```diff
@@ -32,8 +33,8 @@ func (s *APIAdminUsers) AddPaymailToUser(c *gin.Context, id string) {
 	newPaymail := &paymailsmodels.NewPaymail{
 		Alias:      alias,
 		Domain:     domain,
-		PublicName: requestBody.PublicName,
-		Avatar:     requestBody.AvatarURL,
+		PublicName: lo.IfF(requestBody.PublicName != nil, func() string { return *requestBody.PublicName }).Else(""),
```
dorzepowski, line 36: Proposition for simplifying this is to have function like below:

```go
package lox

import "github.com/samber/lo"

type IfElse[T any] interface {
	Else(result T) T
}

func Elvis[T any](value *T) IfElse[T] {
	return lo.IfF[T](value != nil, func() T {
		return *value
	})
}
```

engine/v2/paymails/paymails_service.go
```diff
@@ -40,6 +40,13 @@ func (s *Service) Create(ctx context.Context, newPaymail *paymailsmodels.NewPaym
 		return nil, spverrors.Newf("user does not exist")
 	}
 
+	if err := newPaymail.ValidateAvatar(); err != nil {
+		return nil, spverrors.Wrapf(err, "invalid avatar url during user creation")
```
chris-4chain, line 44: Shouldn't it be returned as defined error and mappend properly on http-handlers level into user's error, like `422` ?

chris-4chain (approved):

dorzepowski (approved):

actions/v2/admin/users/paymail_add.go
dorzepowski: for now I would leave it as is.

PR #949, opened 2025-03-04
Feat(SPV-1544): admin record transaction for user endpoint

dzolt-4chain:
# Pull Request Checklist


actions/v2/admin/transactions/server.go
```diff
@@ -0,0 +1,20 @@
...
+// APIAdminTransactions represents server with admin API endpoints
+type APIAdminTransactions struct {
+	engine engine.ClientInterface
+	logger *zerolog.Logger
+}
+
+// NewAPIAdminTransactions creates a new APIAdminTransactions
+func NewAPIAdminTransactions(engine engine.ClientInterface, logger *zerolog.Logger) APIAdminTransactions {
+	return APIAdminTransactions{
+		engine: engine,
```
dorzepowski, line 17: Please change according to last requirements (so, extract interfaces and assigne only what is needed, not whole engine)

dorzepowski (changes requested):

actions/v2/admin/transactions/record_test.go
```diff
@@ -0,0 +1,619 @@
...
+				"id": "{{ .outpoint }}",
+				"blob": "{{ .blob }}"
+			}`, map[string]any{
+			"outpoint": outpoint.String(),
+			"blob":     dataOfOpReturnTx,
+		})
+	})
+}
+
+func TestDoubleSpending(t *testing.T) {
```
chris-4chain, line 141: I would skip error cases for admin-endpoints because since it is handled by the same engine method it is already tested by other tests anyway.

actions/v2/admin/transactions/record_test.go
```diff
@@ -0,0 +1,619 @@
...
+		then.User(recipient).Operations().Last().
+			WithTxID(txSpec.ID()).
+			WithTxStatus("BROADCASTED").
+			WithValue(1000).
+			WithType("incoming").
+			WithCounterparty(sender.DefaultPaymail().Address())
+	})
+}
+
+func TestTransactionWithChange(t *testing.T) {
```
chris-4chain, line 443: I would skip this test since it is already covered by the one for "user"

actions/v2/admin/transactions/record_test.go
```diff
@@ -0,0 +1,619 @@
...
+			Post(recordTransactionOutlineForUserURL)
+
+		// then:
+		then.Response(res).
+			HasStatus(400).
+			WithJSONf(apierror.ExpectedJSON("error-utxo-spent", "UTXO is already spent"))
+	})
+}
+
+func TestExternalOutgoingTransaction(t *testing.T) {
```
chris-4chain, line 233: I would skip one of these: `TestExternalOutgoingTransaction` or `TestInternalOutgoingTransaction`.

api/components/requests.yaml
```diff
@@ -40,6 +40,18 @@ components:
         - alias
         - domain
 
+    RecordTransactionOutlineForUser:
+      allOf:
+        - $ref: "../components/models.yaml#/components/schemas/TransactionOutline"
```
chris-4chain, line 45: My IDE shows an error here. And actually, there is no `TransactionOutline` in the `models.yaml`

actions/v2/admin/transactions/record_test.go
dorzepowski: I wouldn't

actions/v2/admin/transactions/record_test.go
dorzepowski: I wouldn't

actions/v2/admin/transactions/record_test.go
dorzepowski: I wouldn't

actions/v2/admin/transactions/record_test.go
dzolt-4chain: The general idea of these tests was to test the behaviour of admin delete user. They are the same now, but if something changes under the hood they might use different implementation so I think it would be better to have them here even if they are "the same". It doesn't affect any performance that much but keeps us ensured that everything works correctly

actions/v2/admin/transactions/server.go
dzolt-4chain: done

api/components/requests.yaml
dzolt-4chain: done

